PR #17, opened 2015-11-23
Misc. cleanup and minor features

mdavidsaver:
A collection of unrelated patches, mostly code cleanup.  Two minor feature additions:

1) Add Configuration::getPropertyAsAddress() to parse a property string into osiSockAddr.
2) When built against Base >=3.15.1 AtomicValue uses actual atomic operations.


msekoranja:
3.14.12.5 builds seems to be failing.
I guess this is due the AtomicValue class in
https://github.com/epics-base/pvAccessCPP/blob/master/src/remote/codec.h
and the new introduces one. Should AtomicValue be removed?

mdavidsaver:
Passes now

PR #18, opened 2015-12-09
Specify pvRequest sent when creating ChannelRPC

dhickin:
Add the ability to specify the pvRequest PVStructure sent when creating a Channel RPC using the pvAccess rpcClient library.

This is required for a Diamond middle-layer project where the channel name corresponds to a device. The operation for that device would then be specified when creating the ChannelRPC and the arguments for that operation in the ChannelRPC request itself.

RPCClient::create() is overloaded to allow the pvRequest to be specified whilst retaining the old default for convenience and backwards compatibility.

I have changed the signature of RPCClient's protected constructor. In theory this breaks absolute strict backwards compatibility, although I'm not aware of anyone inheriting from RPCClient. I can provide an overload of the constructor instead if this would be preferred.

### DIFF
--- a/src/rpcClient/rpcClient.cpp
+++ b/src/rpcClient/rpcClient.cpp
@@ -44,9 +44,14 @@ class ChannelAndRPCRequesterImpl :
     Status m_status;
     PVStructure::shared_pointer m_response;
     ChannelRPC::shared_pointer m_channelRPC;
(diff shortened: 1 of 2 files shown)


ralphlange:
retest this please

ralphlange:
retest this please

PR #19, opened 2016-01-06
pvAccessCPP.html

mrkraimer:
The main change is to pvAccessCPP.html.
It now documents pvAccessCPP instead of just directing the reader to pvAccessJava.

I also changed Doxyfile so that generated html goes into directory html instead of documentation/html

I also made changes to pvAccessCPP/src/ioc. Both for ClientRegister and ServerRegister.


ralphlange:
Why did you change the target for the generated HTML?
Was that something we agreed to do?

mrkraimer:
On 01/06/2016 11:33 AM, Ralph Lange wrote:
>
> Why did you change the target for the generated HTML?
> Was that something we agreed to do?
>

Not sure we agreed to it but, as far as I know,  the default location 
for both C++ and Java is a top level html directory.
I think I am the one responsible for making the location documentation/html.
This was a bad idea.

At the APS meeting we did discuss the use of a parent pom.xml files for 
Java,
It was this discussion that lead me to realizes that the major change 
(maybe the only one) that I made to whoever created the original pom.xml
and Doxygen files was to override the location of the generated html 
documentation.

Marty

ralphlange:
Under Maven, the default location for generated Javadoc HTML is `target/site/apidocs`.

Such a location move has a bunch of side-effects for the jenkins scripts (part of the module source tree) and the jenkins jobs (created from yaml templates in the jenkins-jobs repo). Who is going to do the necessary changes?

The jenkins jobs are generated from templates. If single modules change locations of things that have to be deployed, new templates have to be generated, and new macros added for their instantiation.
If Jenkins should be able to re-build older module versions, all affected jobs have to be created in two flavors: old doc location and new doc location.

I would like to see a cost-benefit analysis for such a change.
In any case, I don't think this move should be an individual decision for an individual module.

dhickin:
<p>I don't see the advantage of changing the location of doxygen output. The top-level is already too cluttered.</p>

<p>Personally I would put as little as possible in terms of documentation at top-level, just the README and License and Copyright, with the (short) README pointing to locations of other files if necessary.</p>

mrkraimer:
On 01/07/2016 08:12 AM, dhickin wrote:
>
> I don't see the advantage of changing the location of doxygen output. 
> The top-level is already too cluttered.
>
> Personally I would put as little as possible in terms of documentation 
> at top-level, just the README and License and Copyright, with the 
> (short) README pointing to locations of other files if necessary.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/epics-base/pvAccessCPP/pull/19#issuecomment-169659995>.
>
OK I reverted to HTML_OUTPUT = documentation/html

PR #20, opened 2016-01-25
getSubField -> getSubFieldT to avoid potential NULL de-ref.

mdavidsaver:
Fix a number of instances where the return values of getSubField() is dereferenced w/o checking for NULL.  These were only the most obvious, I suspect there are others.  I also didn't fix those in testApp.

Same issue as epics-base/pvDataCPP#21


mdavidsaver:
shouldn't have assigned myself here.  force of habit

PR #21, opened 2016-01-26
Set SONAME

mdavidsaver:
Set SONAME for libpvAccess using the module version.  Move the module version from pvaVersion.h to src/Makefile.  Now generate pvaVersionNum.h so the version only appears in one place.


src/Makefile
```diff
@@ -33,3 +47,7 @@ pvAccess_SYS_LIBS_WIN32 += ws2_32
 
 include $(TOP)/configure/RULES
 
+# Can't use EXPAND as generated headers must appear
+# in O.Common, but EXPAND emits rules for O.$(T_A)
+../O.Common/pvaVersionNum.h: pvaVersionNum.h@
+	$(EXPAND_TOOL) $(EXPANDFLAGS) $($@_EXPANDFLAGS) $< $@
```
mdavidsaver: @anjohnson Can you suggest some way to avoid a custom rule here?  I couldn't see an obvious way to get $(EXPAND) to install to O.Common.

msekoranja:
@mdavidsaver Build fails:

make[2]: *** No rule to make target `pvaVersionNum.h@', needed by `../O.Common/pvaVersionNum.h'.  Stop.

anjohnson:
@msekoranja why did you merge this so soon, especially as it appears to have already had a build failure showing? Please revert the merge so we can fix it first.

Quoting from our [github pull request guidelines](https://gist.github.com/anjohnson/544591e671b03f1a3131):
* Our group members are spread out over many different countries and timezones. Everyone should get at minimum a full working day to respond to a pull request, allowing for weekends and local public holidays.

mdavidsaver:
Ah, the failure is probably to a vpath directive which was added to RULES_EXPAND in 3.15

> vpath %@ $(USR_VPATH) $(ALL_SRC_DIRS)

I've pushed another change to this branch (771663f42142cfcc524e3434671cb00eb3b63d0e) which uses the full relative path, which should work across all base versions.

mdavidsaver:
Since @msekoranja is probably out enjoying his evening I've just pushed the fix.  The whole thing can be reverted if necessary.

src/Makefile
anjohnson: @mdavidsaver I played around with this a bit but with no joy. You don't need to set EXPANDFLAGS, that's done for you already in base/configure/RULES_EXPAND since you're using the same variable names. The rule itself should really appear in a new RULES snippet file in the src/pva sub-directory, which would need including from the bottom of the parent Makefile, but that's a minor point that can be fixed later.

mdavidsaver: > I played around with this a bit but with no joy. 

Well no surprises there.  It is after all a Makefile :)

> You don't need to set EXPANDFLAGS, that's done for you already in base/configure/RULES_EXPAND since you're using the same variable names.

There is a subtle difference in definitions.  I have to say I find

> EXPAND_VARS += EPICS_PVA_MAJOR_VERSION="$(EPICS_PVA_MAJOR_VERSION)"

a little verbose.

> The rule itself should really appear in a new RULES snippet file in the src/pva sub-directory, which would need including from the bottom of the parent Makefile, but that's a minor point that can be fixed later.

Fair enough.  Just laziness or my part.

anjohnson: > There is a subtle difference in definitions. I have to say I find
> 
>     EXPAND_VARS += EPICS_PVA_MAJOR_VERSION="$(EPICS_PVA_MAJOR_VERSION)"
> 
> a little verbose.

Ah, didn't spot that little refinement. Of course you could have used
shorter variable names in the local Makefile, say MAJOR, MINOR, MAINT
and, er, DEVMT say.

* Andrew

-- 
There are only two hard problems in distributed systems:
  2. Exactly-once delivery
  1. Guaranteed order of messages
  2. Exactly-once delivery
 -- Mathias Verraes

msekoranja:
@anjohnson I naively assumed Michael resolved the issue before requesting for a pull request.

PR #25, opened 2016-02-10
remove use of throw() specifier

mdavidsaver:
Function throw() specifiers in C++ have never been useful, and are deprecated in c++11.  In my mind almost[1] any use of this "feature" is a bug as the author probably did not understand the large difference in behavior between between c++ and java.  In java these are used for compile time checks, in c++ for runtime checks (like assert() complete with abort()).

[1] The "almost" is the use of throw() to indicate that no exception should ever be thrown.

### DIFF
--- a/documentation/pvAccessCPP.html
+++ b/documentation/pvAccessCPP.html
@@ -1370,7 +1370,7 @@ class  RPCService
     virtual ~RPCService() {};
     virtual PVStructurePtr request(
        PVStructurePtr const &amp; args
-    ) throw (RPCRequestException) = 0;
+    ) = 0;
 };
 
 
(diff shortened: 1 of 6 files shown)


anjohnson:
I agree with the point about not using throw specifiers, except on APIs that are designed to be called from C code or similar situations where throw should never be used. However there are build check failures with this specific pull request:
```
.../lib/linux-x86_64/libpvAccess.so: undefined reference to `vtable for epics::pvAccess::SecuritySession'
.../lib/linux-x86_64/libpvAccess.so: undefined reference to `typeinfo for epics::pvAccess::SecuritySession'
```

mdavidsaver:
> However there are build check failures

oops, removed a '=0' by accident.

mdavidsaver:
> ... situations where throw should never be used ...

To be exact, this doesn't mean that 'throw' should not be used in the function.  Just that it must be caught so that it doesn't propagate out of the function (as this would result in abort() being called).

So use of 'throw()' means either careful auditing of the function body to ensure that nothing would throw, or wrapping the entire function body in a try{} block.

PR #26, opened 2016-02-27
bugfix for segfault when using pvaClientCPP

klemenv:
This patch fixes the problem we've been seeing at SNS when pvaPy is using pvaClientCPP and eventually crashes. When running out of free pre-allocated pvStructures, the last one is used to somehow keep track of missed changes. But it must not be handed to user since it's modified by pvAccess whenever there's data, until there's at least one free pvStructure.

### DIFF
--- a/src/remoteClient/clientContextImpl.cpp
+++ b/src/remoteClient/clientContextImpl.cpp
@@ -2204,6 +2204,7 @@ namespace epics {
                    {
                        m_overrunInProgress = true;
                        m_overrunElement = newElement;
+                       return;
                    }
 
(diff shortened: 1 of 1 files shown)


ralphlange:
ok to test

msekoranja:
The bugfix is not correct, i.e. the response is ignore and monitors lost. The response needs to be deserialized and not put to the queue (as it is now).

mdavidsaver:
For the record, Matej's fix was fea2e1c26390158ec1a93c14abcde58a6ebc8101

PR #34, opened 2016-03-18
Channel defaults

mdavidsaver:
Provide and use a default implementation for some Channel methods.  Default Requester methods forward to associated ChannelRequester, default create*() methods which fail with "Not Implemented", and default no-op printInfo().

The aim is to reduce the amount of boilerplate someone writing a new ChannelProvider has to add.

### DIFF
--- a/src/ca/caChannel.cpp
+++ b/src/ca/caChannel.cpp
@@ -277,13 +277,6 @@ std::tr1::shared_ptr<ChannelRequester> CAChannel::getChannelRequester()
     return channelRequester;
 }
 
-
-bool CAChannel::isConnected()
-{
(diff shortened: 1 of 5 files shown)


msekoranja:
@mdavidsaver Why not using predefined status Status-es instead of creating (although on stack) new one?

mdavidsaver:
> Why not using predefined status Status-es instead of creating (although on stack) new one?

It introduces potential initialization order fun for no good reason.  Constructing empty std::string is cheap (no allocation).

msekoranja:
That's true... also for all static variables. Since performance does not play any role here I will merge.

PR #36, opened 2016-03-20
Add requester_type and operation_type

mdavidsaver:
Typedefs to link between *Requester classes and associated "operation"s.  eg. ```ChannelGet::requester_type``` is ```ChannelGetRequester``` and ```ChannelGetRequester::operation_type``` is ```ChannelGet```.

ChannelFind/List is an ambiguous.  All others are 1-to-1.  ```ChannelFind::requester_type``` is ```ChannelFind``` while both ```ChannelListRequester::operation_type``` and ```ChannelFindRequester::operation_type``` are ```ChannelFind```.

### DIFF
--- a/src/client/pv/pvAccess.h
+++ b/src/client/pv/pvAccess.h
@@ -108,6 +108,13 @@ private:
 
 class Channel;
 class ChannelProvider;
+class ChannelArrayRequester;
+class ChannelFindRequester;
+class ChannelGetRequester;
+class ChannelProcessRequester;
+class ChannelPutRequester;
+class ChannelPutGetRequester;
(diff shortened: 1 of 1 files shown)


msekoranja:
I see this as template enabler feature. I like it.
Since one anomaly could break whole thing I am asking: does it makes sense to introduce ChannelList and make "1-to-1 mapping" mentioned above.

mdavidsaver:
> ... does it makes sense to introduce ChannelList ...

I wouldn't hurt, but I don't think it's that important.

msekoranja:
Since this would be an API change I would not do it for now (can be done if needed), and will merge as requested.

PR #46, opened 2016-04-19
PVAServerRegister:: allow providerList, do not use epicsAtExit

mrkraimer:
Title describes the change.

### DIFF
--- a/src/ioc/PVAClientRegister.cpp
+++ b/src/ioc/PVAClientRegister.cpp
@@ -32,10 +32,6 @@ using std::endl;
 using namespace epics::pvData;
 using namespace epics::pvAccess;
 
-//static void pvaClientExitHandler(void* /*pPrivate*/) {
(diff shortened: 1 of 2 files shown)


anjohnson:
Why is there so much commented-out code here? You can always retrieve it from git if you're not sure about deleting and you later find that you did need it; much better to delete it so it doesn't hang around.

mrkraimer:
On 04/19/2016 11:32 AM, Andrew Johnson wrote:
>
> Why is there so much commented-out code here? You can always retrieve 
> it from git if you're not sure about deleting and you later find that 
> you did need it; much better to delete it so it doesn't hang around.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub 
> <https://github.com/epics-base/pvAccessCPP/pull/46#issuecomment-211980027>
>
OK I removed them.

PR #47, opened 2016-04-21
Improve test robustness (was failing on a Mac VM)

anjohnson:
Store and compare the complete time-stamp, not just the seconds field.

### DIFF
--- a/testApp/remote/channelAccessIFTest.cpp
+++ b/testApp/remote/channelAccessIFTest.cpp
@@ -632,11 +632,12 @@ void ChannelAccessIFTest::test_channelGetIntProcessInternal(Channel::shared_poin
         }
 
         pvTimeStamp.get(timeStamp);
+        double deltaT = TimeStamp::diff(timeStamp, previousTimestamp);
 
         testOk((previousValue +1)/*%11*/ == value->get(), "%s: testing the counter value change",
                testMethodName.c_str());
-        testOk(timeStamp > previousTimestamp,
-               "%s: testing the timestamp change", testMethodName.c_str());
+        testOk(deltaT > 0.9 && deltaT < 2.0,
+               "%s: timestamp change was %g", testMethodName.c_str(), deltaT);
     }
 
     channelGetReq->getChannelGet()->destroy();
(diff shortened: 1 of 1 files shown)


msekoranja:
Thanks.
This fixed does fix the test, but does remove a check whether the timestamp increased for ~ 1 second. That was not done robust enough in current code.

Maybe something like (not sure wether epicsTimeStamp can do this):

(lastTimeStamp - timeStamp) > 0.9 && 
(lastTimeStamp - timeStamp) < 1.5

would be better.

anjohnson:
Whoever wrote the V4 TimeStamp class (in pvData pv/timeStamp.h) that this uses (it's not an epicsTime) did not provide the standard math operators for it, although it does have a diff() function. There is no strftime() method or operator<< either, which would be pretty useful.

I had to make the test (deltaT > 0.9 && deltaT < 2.0) since one loop took over 1.5 seconds on my (pretty fast) workstation; hope it doesn't take even longer on some of my build slaves.

Please take another look.

msekoranja:
Excellent!

mrkraimer:
On 04/21/2016 04:35 PM, Andrew Johnson wrote:
>
> Whoever wrote the V4 TimeStamp class (in pvData pv/timeStamp.h) that 
> this uses (it's not an epicsTime) did not provide the standard math 
> operators for it, although it does have a diff() function.
>

It does has a lot of the standard operators, e. g. ==, <=, etc
>
> There is no strftime() method or operator<< either, which would be 
> pretty useful.
>

Agreed that these would be useful.

> I had to make the test (deltaT > 0.9 && deltaT < 2.0) since one loop 
> took over 1.5 seconds on my (pretty fast) workstation; hope it doesn't 
> take even longer on some of my build slaves.
>

What was taking so long?
It should not be diff since it's implementation is:

double TimeStamp::diff(TimeStamp const & a,TimeStamp const & b)
{
     double result = static_cast<double>(a.secondsPastEpoch - 
b.secondsPastEpoch);
     result += (a.nanoseconds - b.nanoseconds)/1e9;
     return result;
}

anjohnson:
TimeStamp could implement these member functions to make it more like a C++ value type:
``` C++
double TimeStamp::operator- (const TimeStamp &that);
TimeStamp TimeStamp::operator+ (const double &secs);
TimeStamp TimeStamp::operator- (const double &secs);
```
so that we could write these kinds of natural expressions:
``` C++
double deltaT = timeStamp - previousTimestamp;
TimeStamp nextTick = timeStamp + 0.5;
```
The `epicsTime` class has many more constructors, which implement conversion functions from other common ways in which time is represented in software.

It is also possible to implement this non-member function, although less important than the above:
``` C++
TimeStamp operator+ (const double &secs, const TimeStamp &that);
```
which allows this
``` C++
TimeStamp nextTick = 0.5 + timeStamp;
```

PR #56, opened 2016-07-18
Correctly set status in caProvider

dhickin:


dhickin:
3.14 and 3.16 build are OK but 3.15 PR build fails because it's trying to use a mixture of 3.15.3 and 3.15.4:

<a href="https://openepics.ci.cloudbees.com/job/pvAccessCPP_PullRequests/38/BASE=3.15.3,USE_MB=MB_NO/console">3.15 console</a>

This is issue with jenkins/cloudbees. I don't believe there's an issue with the code.

PR #57, opened 2016-07-18
caProvider: Remove control/valueAlarm for arrays

dhickin:
This brings caProvider behaviour in line with pvaSrv.

See discussion on epics-base/pvaSrv/pull/38 and the corresponding pull request epics-base/pvaSrv/pull/40

### DIFF
--- a/src/ca/caChannel.cpp
+++ b/src/ca/caChannel.cpp
@@ -191,17 +191,28 @@ static PVStructure::shared_pointer createPVStructure(CAChannel::shared_pointer c
     // Match to closest DBR type
     // NOTE: value is always there
     string properties;
+    bool isArray = channel->getElementCount() > 1;
     if (dbrType >= DBR_CTRL_STRING)      // 28
     {
         if (dbrType != DBR_CTRL_STRING && dbrType != DBR_CTRL_ENUM)
-            properties = "value,alarm,display,valueAlarm,control";
+        {
+            if (isArray)
+                properties = "value,alarm,display";
+            else
+                properties = "value,alarm,display,valueAlarm,control";
+        }
         else
             properties = "value,alarm";
     }
     else if (dbrType >= DBR_GR_STRING)   // 21
     {
         if (dbrType != DBR_GR_STRING && dbrType != DBR_GR_ENUM)
-            properties = "value,alarm,display,valueAlarm";
+        {
+            if (isArray)
+                properties = "value,alarm,display";
+            else
+                properties = "value,alarm,display,valueAlarm";
+        }
         else
             properties = "value,alarm";
     }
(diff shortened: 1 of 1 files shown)


dhickin:
3.15 build failed due to mixing 3.15.3 and 3.15.4 again. Otherwise OK.

msekoranja:
Why aren't control/valueAlarm relevant for arrays? They can apply for array elements?

dhickin:
1. Note while NTScalar has optional control and alarmLimit (the equivalent to valueAlarm) fields, NTScalarArray doesn't. They can be added as additional fields but their omission from NTScalarArray suggests  omitting them from the returned structure. (Otherwise, why are they not part of NTScalarArray?)

2. What is the meaning of control and alarm for arrays? Does this imply that we check each element of the array against the control limits and alarm limits. Do we raise an alarm if one value is outside the alarm limits? Do we check each element and limit it if it is outside control limits?

3. In practical terms

   a) do implementations do the behaviour described in 2?

   b) Are these values filled in with anything other than defaults (e.g. for the control fields, display values for VAL fields or max/min for the numeric type)? For the examples in testTop the waveform records produce through caProvider control limits which mirror the display limits and no meaningful alarm limits. They don't for example have DRVL/H fields like the scalars. Nor do they have alarm fields like HHSV, LOLO or HYST. 

4. Whatever we do for caProvider we need to do for pvaSrv and vice versa, both in terms of structures and values. For pvaSrv the control and valueAlarm fields were removed from array PVs.

I spent a fair bit of time getting these in line - these current pull requests are the last step in getting everything the same. This was discussed in the pvaSrv pull requests and presented at the ESS meeting.

However if we want the control and valueAlarm fields for numeric arrays for caProvider then I guess we can revert the pvaSrv structures to match caProvider and then get the values the same.

mrkraimer:
On 07/19/2016 09:59 AM, dhickin wrote:
>
> 1.
>
>     Note while NTScalar has optional control and alarmLimit (the
>     equivalent to valueAlarm) fields, NTScalarArray doesn't. They can
>     be added as additional fields but their omission from
>     NTScalarArray suggests omitting them from the returned structure.
>     (Otherwise, why are they not part of NTScalarArray?)
>
> 2.
>
>     What is the meaning of control and alarm for arrays? Does this
>     imply that we check each element of the array against the control
>     limits and alarm limits. Do we raise an alarm if one value is
>     outside the alarm limits? Do we check each element and limit it if
>     it is outside control limits?
>
> 3.
>
>     In practical terms
>
>     a) do implementations do the behaviour described in 2?
>
>     b) Are these values filled in with anything other than defaults
>     (e.g. for the control fields, display values for VAL fields or
>     max/min for the numeric type)? For the examples in testTop the
>     waveform records produce through caProvider control limits which
>     mirror the display limits and no meaningful alarm limits. They
>     don't for example have DRVL/H fields like the scalars. Nor do they
>     have alarm fields like HHSV, LOLO or HYST.
>
> 4.
>
>     Whatever we do for caProvider we need to do for pvaSrv and vice
>     versa, both in terms of structures and values. For pvaSrv the
>     control and valueAlarm fields were removed from array PVs.
>
> I spent a fair bit of time getting these in line - these current pull 
> requests are the last step in getting everything the same. This was 
> discussed in the pvaSrv pull requests and presented at the ESS meeting.
>
> However if we want the control and valueAlarm fields for numeric 
> arrays for caProvider then I guess we can revert the pvaSrv structures 
> to match caProvider and then get the values the same.
>

For the reasons David lists above,I think that array value fields should 
NOT, by default, have control and valueAlarm fields.

Note that in epics base the waveform record defines control fields but 
not value alarm fields.
But the record support has no code to support control limits.
Device support could, but I will guess that this has almost never been used.

Marty

msekoranja:
It seems that this issue was already discussed at the ESS meeting, and agreed.

PR #60, opened 2017-04-08
Test assumptions about ChannelProvider ownership

mdavidsaver:
Add testProvider unittest as a work in progress to document/tests my assumptions about the ownership of Channel and ChannelGet returned by the PVA and CA client providers.  What I see is that complete shutdown (eg. destroy() provider) does cleanup everything, but that variations on partial shutdown don't seem to.  At least in the ways I'm expecting.

I know that some of these tests should fail now.  This will hopefully change during this discussion.

Specifically, I expect that:

1. Calling ```->destroy()``` on Channel/ChannelGet/... should (immediately?) release references to the corresponding ```Requester```.

2. Calling ```->destroy()``` on Channel/ChannelGet/... should (eventually?) release all internal references.

3. While a ```Channel``` may be shared implicitly, a ```ChannelGet``` or similar operation accepting a pvRequest is exclusively owned by the caller of ```createChannelGet()```.

4. Calling ```->destroy()``` on Channel will implicitly destroy all operations (Channel*).

5. Orphaned Channel instance (no external references) will eventually be closed?

Are these expectations correct?  Apparent violation of 1 by ChannelGet w/ PVA is causing me trouble.

### DIFF
--- a/testApp/remote/Makefile
+++ b/testApp/remote/Makefile
@@ -12,6 +12,9 @@ testCodec_SRCS = testCodec
 testHarness_SRCS += testCodec.cpp
 TESTS += testCodec
 
+TESTPROD_HOST += testProviders
+testProviders_SRCS += testProviders.cpp
+TESTS += testProviders
 
 TESTPROD_HOST += testRemoteClientImpl
 testRemoteClientImpl_SRCS += testRemoteClientImpl.cpp
@@ -21,7 +24,6 @@ testChannelConnect_SRCS += testChannelConnect.cpp
 
 TESTPROD_HOST += testServerContext
 testServerContext_SRCS += testServerContext.cpp
-
 
 
 PROD_HOST += testServer
--- a/testApp/remote/testProviders.cpp
+++ b/testApp/remote/testProviders.cpp
@@ -1,0 +1,209 @@
+/**
+ * Copyright - See the COPYRIGHT that is included with this distribution.
+ * pvAccessCPP is distributed subject to a Software License Agreement found
+ * in file LICENSE that is included with this distribution.
+ */
+
+#include <epicsUnitTest.h>
+#include <testMain.h>
+#include <epicsThread.h>
+
+#include <pv/pvAccess.h>
+#include <pv/clientFactory.h>
+#include <pv/caProvider.h>
+
+namespace {
+
+namespace pvd = epics::pvData;
+namespace pva = epics::pvAccess;
+
+struct TestChanReq : public pva::ChannelRequester
+{
+    POINTER_DEFINITIONS(TestChanReq);
+
+    virtual std::string getRequesterName() { return "TestChanReq"; }
+
+    virtual void channelCreated(const pvd::Status& status, pva::Channel::shared_pointer const & channel)
+    {}
+
+    virtual void channelStateChange(pva::Channel::shared_pointer const & channel, pva::Channel::ConnectionState connectionState)
+    {}
+};
+
+struct TestGetReq : public pva::ChannelGetRequester
+{
+    POINTER_DEFINITIONS(TestGetReq);
+
+    virtual std::string getRequesterName() { return "TestGetReq"; }
+
+    virtual void channelGetConnect(
+        const pvd::Status& status,
+        pva::ChannelGet::shared_pointer const & channelGet,
+        pvd::Structure::const_shared_pointer const & structure)
+    {}
+
+    virtual void getDone(
+        const pvd::Status& status,
+        pva::ChannelGet::shared_pointer const & channelGet,
+        pvd::PVStructure::shared_pointer const & pvStructure,
+        pvd::BitSet::shared_pointer const & bitSet)
+    {}
+};
+
+static
+void testChannelCreateDestroy(const char *providerName,
+                              const char *badChannel)
+{
+    testDiag("testChannelCreateDestroy(\"%s\", \"%s\")", providerName, badChannel);
+
+    pva::ChannelProvider::shared_pointer provider(pva::getChannelProviderRegistry()->createProvider(providerName));
+    pva::ChannelProvider::weak_pointer providerW(provider);
+
+    testOk(provider.unique(), "create provider refs=%u", (unsigned)provider.use_count());
+
+    TestChanReq::shared_pointer req(new TestChanReq);
+    pva::Channel::shared_pointer chan(provider->createChannel(badChannel, req));
+    pva::Channel::weak_pointer chanW(chan);
+
+    testOk(chan.unique(), "create non-existant channel refs=%u", (unsigned)chan.use_count());
+    epicsThreadSleep(1.0); // wait in case some worker thread is lazy...
+    testOk(chan.unique(), "2create non-existant channel refs=%u", (unsigned)chan.use_count());
+
+    chan->destroy();
+
+    testOk(chan.unique(), "destroy non-existant channel refs=%u", (unsigned)chan.use_count());
+    testOk(req.unique(), "ChannelRequester after Channel::destroy() refs=%u", (unsigned)req.use_count());
+    epicsThreadSleep(1.0);
+    testOk(chan.unique(), "2destroy non-existant channel refs=%u", (unsigned)chan.use_count());
+    testOk(req.unique(), "2ChannelRequester after Channel::destroy() refs=%u", (unsigned)req.use_count());
+
+    chan.reset();
+    testOk(req.unique(), "ChannelRequester after Channel last refs=%u", (unsigned)req.use_count());
+    req.reset();
+
+    testOk(!chanW.lock(), "Channel leaked from live Provider");
+
+    testOk(provider.unique(), "empty provider refs=%u", (unsigned)provider.use_count());
+    epicsThreadSleep(1.0);
+    testOk(provider.unique(), "2empty provider refs=%u", (unsigned)provider.use_count());
+
+    provider->destroy();
+
+    testOk(provider.unique(), "empty provider refs=%u", (unsigned)provider.use_count());
+    epicsThreadSleep(1.0);
+    testOk(provider.unique(), "2empty provider refs=%u", (unsigned)provider.use_count());
+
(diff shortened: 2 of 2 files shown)


msekoranja:
I've searched for some email I remember I have written while ago:

ChannelProvider (aka InternalClientContextImpl) holds weak pointers to all the Channels and requests (e.g. ChannelGet).

Channel (aka InternalChannelImpl) holds shared pointer to the ChannelProvider, ChannelRequester and corresponding Transport. There is also a weak pointer to the requests each channel owns.
Transports hold weak pointer to the Channels.

So, if there is an instance of a Channel and there a programmer releases all the references to it, there are actually no references to the Channel itself and its destructor will be called.
This is all OK.

Now, lets check ChannelGet request.

ChannelGetImpl holds a shared pointer to ChannelGetRequester and Channel.

So, a Channel cannot be destroyed until it has active ChannelGet request (example of "active": get() method is called and until response is received the request is "active"). That’s as expected. Also there is no shared pointer (only weak pointers) to the ChannelGetImpl.
If programmer holds no reference to ChannelGetImpl its destroctor is being called. This will release shared pointer to the Channel. And Channel get be destroyed afterwards.

(NOTE: my design approach is: a child (aka Channel) always holds a shared_ptr to the parent (aka ChannelProvider) since it cannot exist without the parent. A parent holds only a weak_ptr to its children.)

----

All your expectations are correct, except 1) and maybe 2). 
If you are 100% sure that code will use Requester's shared_ptr after destroy, then the pointer can be reset. 
Anyway, the Requesters should never hold a shared_ptr to "requestees" to avoid cyclic shared_ptr dependency. Do you want to call destroy() to break these?

mdavidsaver:
To rephrase as a test cases:

> So, if there is an instance of a Channel and there a programmer releases all the references to it, there are actually no references to the Channel itself and its destructor will be called.

So createChannel() returns a shared_ptr which is unique()==true

> provider->createChannel("blah...", chReq).unique()==true


> If programmer holds no reference to ChannelGetImpl its destroctor is being called.

And the same for createChannelGet()

> chan->createChannelGet(pvReq, chGetReq).unique()==true


Correct?

mdavidsaver:
> Anyway, the Requesters should never hold a shared_ptr to "requestees" to avoid cyclic shared_ptr dependency.

I can see this, however I find that avoiding it is awkward.

> Do you want to call destroy() to break these?

Partly, this.  Also I was hoping to use this to cancel an operation.

msekoranja:
> From my tests, I see that from the PVA client provider, createChannel() and createChannelGet() return a non-unique shared_ptr (ref count=2). The CA provider returns unique() shared_ptr.

> From a quick look, the second Channel ref is in SimpleChannelSearchManagerImpl::m_channels.

You are right. When channel is found then the ref. count will drop to 1 (as expected).
SimpleChannelSearchManagerImpl::m_channels should use weak_ptr (and code adopted to it - to handle lock() == 0 cases)!

> For ChannelGet, the second ref appears to be BaseRequestImpl::m_thisPointer which looks like ref. loop.

That's OK.
Imagine the following full async code (where returned ChannelGet::shared_ptr is not saved):

channel->createChannelGet(myChannelGetRequester, pvRequest);

This will result in immediate destruction of ChannelGet instance, since ChannelGet::shared_ptr will drop to 0. I use BaseRequestImpl::m_thisPointer to hold the pointer to self until the request is pending (e.g. until response is received locally). m_thisPointer is released when response from the server responds.

mdavidsaver:
> Imagine the following full async code (where returned ChannelGet::shared_ptr is not saved):

I see what you're trying to achieve, and why.  This is exceedingly subtle, as evidenced by the fact that the CA provider doesn't do this.  I've added a new test ```testChannelDrop()``` to demonstrate this.  Neither do the Providers I've written for pva2pva.  This means that the "weak" ref stored in Channel and Context are functionally strong refs until the first reply is received.

So my expectation #3 doesn't hold either.  (I'm not doing so good here...)

IMO uniformity among Providers is more important that supporting this fully async case.  It might be simplest to not support this case as you describe it, otherwise I think some change to the ownership rules or API is called for.

For example, reverse the expected ownership between Requester and Operation.  Currently Operation has a strong ref. to Requester.  Flip this so that Operation has a weak ref to Requester, which will allow the Requester to have a strong ref to the Operation.  It seems to me that an Operation can always tolerate finding that its Requester has been destroy'd.

Then the fully async case becomes something like:

```
struct MyGetReq : public pva::ChannelGetRequester {
   ...
   pva::ChannelGet::shared_pointer op;
};
{
  ...
  shared_ptr<MyGetReq> req(new MyGetReq);
  pva::ChannelGet::shared_pointer op = channel->createChannelGet(pvRequest, req);
  {
    epicsMutexGuard G(....);
    // use a mutex here and in connectChannelGet() to prevent race when assigning MyGetReq::op
    req->op = op;
  }
  ...
```

I'll be thinking some more about this...

mdavidsaver:
> Anyway, the Requesters should never hold a shared_ptr to "requestees" to avoid cyclic shared_ptr dependency.

The two non-trivial* c++ usages of client ChannelGet I can find (and my pvalink with Monitor) are apparently violating this rule and hopefully calling destroy() to avoid leaks.  By "trivial" I mean using callback for something other than to call epicsEvent::signal() or similar.

Interestingly, ADNeD is also in the waitForConnect() category, and does all work in monitorEvent().  (fine as long as only one operation per channel)

mdavidsaver:
Another inconsistency between PVA and CA providers is that the PVA provider can create Operations on a disconnected Channel, but CA can't.  Client code I've written doesn't assume this, but I notice that at least pvPortClient does.

mdavidsaver:
Ok, I see.  pvAccessJava allows the "fully async" behavior as the Channel keeps strong refs to all operations, which is a ref loop, but apparently the GC can untie the knot.

Java code can play fast and loose with ownership in this way.  IMO this should not be attempted w/ c++.  I'm then led to the conclusion that the fully async use case isn't safe in c++.  I also can't find any c++ code attempting it now.

From this, I see m_thisPointer as a bug.  As far as I can tell, removing it will not break any existing, public, code.  I'd mark createChannel* with gcc attribute must_check in case I missed something.

@msekoranja will you agree with removing m_thisPointer?  If so do have interest/time to make this change or should I?

msekoranja:
I find "full async" crucial and very important. This is why m_thisPointer is needed, and is used only when request is "active". I would rather add missing capabilities to other providers instead of removing this feature. There is a testApp/remote/channelAccessIFTest.cpp test that (fully) tests the providers. This should be always updated and tested against all the providers. If a provider passes that test, then it is "approved".

Note: storing (with necessary sync) "pva::ChannelGet::shared_pointer op" is usually not necessary, since all the callbacks have the "op" as argument.

mdavidsaver:
> I find "full async" crucial and very important.

You and I may have to agree to dis-agree on this point.  On reflection, what I don't like is that this leaves no clear owner of the Operation until the Connect method is called.  This may never happen.  So this pattern provides no means to cancel/abort in progress operations on disconnected channels!  I regard being able to cleanup some, but not all, operations and channels as critical for any long lived client (eg. gateway, archiver, or opi).

I've made a search of publicly accessible code using pvAccess (both java and c++) and can find no c++ code which attempts the "fully async" pattern.  I can see that much java code does.  My statements above apply to java as well, so I'm disinclined to consider this good practice.

> Note: storing (with necessary sync) "pva::ChannelGet::shared_pointer op" is usually not necessary, since all the callbacks have the "op" as argument.

I need to track in-progress operations at all stages so that I could cancel them.

msekoranja:
"Progress operations" will always be cleaned-up on disconnected channel. Transition to disconnected state will automatically cancel all the operations (and release m_thisPointer).

I've looked at the first code, e.g. pvget

https://github.com/epics-base/pvAccessCPP/blob/master/pvtoolsSrc/pvget.cpp#L627

 The code intentionally stores "ChannelGet::shared_pointer channelGet" reference. But there is no need to. If I remove this store, then I get the async pattern described above.
A big problem I see if that without m_thisPointer everything cleans-up without any notification. 95% of programmers will never figure it why.

I am not insisting that m_thisPointer should stay or that there might be a better/simpler solution. We just need to find a clean/nice one.

ps I have no time to do any work on this.

mdavidsaver:
> "Progress operations" will always be cleaned-up on disconnected channel. Transition to disconnected state will automatically cancel all the operations (and release m_thisPointer).

Right, true.  However, by "disconnected channels" I also include channels which are never connected.  These will simply accumulate operations until they become connected.

With this in mind, I find myself looking skeptically at code like

https://github.com/ControlSystemStudio/cs-studio/blob/0a1e8699e24808bf7642ed21cbdf29b39f0e5dc6/core/base/base-plugins/org.csstudio.vtype.pv/src/org/csstudio/vtype/pv/pva/PVA_PV.java#L291

which appears to be blindly starting ChannelPut operations without checking for connection state, or for previous put operations.  Seems like a recipe for a burst of old puts on (re)connect.

> 95% of programmers will never figure it why.

This is straightforward.  Put a print in the dtor.  imo, with c++ when something returns shared_ptr<> I assume I'm meant to hold onto this.

More confusing is wondering why objects continue to live after I think I've dropped all references.  I can't speak for X% of programmers, but I find this difficult most difficult to troubleshoot.

msekoranja:
Yes, there is a "feature" on disconnected (and never connected) channels to store request and wait until channel is connected. Usually, this quite simplifies client code. User can always check if channel state first (with a chance of a race condition - API should have a param to allow/disallow that).

Note that you are (1-X)% of programmers, where X is dangerously close to 1 :)
A agree with your last sentence. Options: do not hold references internally, or document the cases when this happens.

mdavidsaver:
Another item on the topic of PVA vs. CA provider differences.  The CA provider doesn't implement a monitor queue and calls monitorEvent each time.  So the single PVStructure instance can concurrently be written from the CA event callback, and read from user code.  Race conditions abound...

mdavidsaver:
This PR has served it's purpose.

PR #67, opened 2017-09-07
caProvider: use weak_pointer where required

mrkraimer:
Use weak_pointer where required.
All destroy methods are  now deprecated and do nothing.


mrkraimer:
It is now 22 days since I created the pull request.
It is about 10 days since Michael said he would look at changes.
So when can the pull request be merged?
Marty

sveseli:
This PR cleans up code significantly and is ok to merge. There are still some problems that can be seen when running valgrind against example clients that involve caProvider in exampleCPP, but those likely do not originate in caProvider itself.

PR #70, opened 2017-11-13
remove lastRequest messages; in activate only make callback if connected

mrkraimer:
remove message from lastRequest both put and ger

In actiivate (channelGet, channelPut, and monitor) call user callback only if channel is connected.

Add more debug statements

### DIFF
--- a/src/ca/caChannel.cpp
+++ b/src/ca/caChannel.cpp
@@ -672,7 +672,7 @@ CAChannelGet::CAChannelGet(CAChannel::shared_pointer const & channel,
   :
     channel(channel),
     channelGetRequester(channelGetRequester),
-    pvRequest(pvRequest)
+    pvRequestWeak(pvRequest)
 {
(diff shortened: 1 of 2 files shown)


anjohnson:
In email Siniša posed a question about the additional use of `if (DEBUG_LEVEL>0)` which does add quite a few lines to this pull request and might better have been included in a separate commit for ease of review (Marty if you haven't tried it yet you might want to play with the `git gui` command; its right-click menu lets you select particular hunks of changes and even individual lines to be included or excluded from a commit, making it easy to group and describe sets of modifications that belong together logically even when the changes span multiple files and only cover some of the changes to the individual files. It's very different to the way we worked in CVS...).

However given that the CA provider code already uses this approach to debugging I don't have a problem with merging this as is, but we might want to look at unifying how we do debugging across the pvAccess module in the future. `git grep -i debug` shows many different debugging styles in use just across this one module (but then the same is probably true for Base itself).

PR #79, opened 2017-12-23
When caProvider is destroyed make sure all channels are cleared

mrkraimer:
Make sure that when CACProvider is destroyed that ca_clear_channel has been called for all channels.

### DIFF
--- a/src/ca/caProvider.cpp
+++ b/src/ca/caProvider.cpp
@@ -52,7 +52,28 @@ CAChannelProvider::CAChannelProvider(const std::tr1::shared_ptr<Configuration>&)
 
 CAChannelProvider::~CAChannelProvider()
 {
-    if(DEBUG_LEVEL>0) std::cout << "CAChannelProvider::~CAChannelProvider()\n";
+    if(DEBUG_LEVEL>0) {
+         std::cout << "CAChannelProvider::~CAChannelProvider()"
+            << " caChannelList.size() " << caChannelList.size()
+            << std::endl;
+    }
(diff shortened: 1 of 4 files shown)


src/ca/caProvider.cpp
```diff
@@ -154,7 +190,9 @@ void ca_factory_cleanup(void*)
         ChannelProviderRegistry::clients()->remove("ca");
         ca_context_destroy(); 
     } catch(std::exception& e) {
-        LOG(logLevelWarn, "Error when unregister \"ca\" factory");
+        std::string message("Error when unregister \"ca\" factory");
+        message += e.what();
+        LOG(logLevelWarn,message.c_str());
```
anjohnson: This `LOG()` macro gives me a compiler warning:
```
In file included from ../caProvider.cpp:13:0:
../caProvider.cpp: In function ‘void epics::pvAccess::ca::ca_factory_cleanup(void*)’:
/home/anj/Software/epics/base-7.0/include/pv/logger.h:48:73: warning: format not a string literal and no format arguments [-Wformat-security]
 #define LOG(level, format, ...) pvAccessLog(level, format, ##__VA_ARGS__)
                                                                         ^
../caProvider.cpp:195:9: note: in expansion of macro ‘LOG’
         LOG(logLevelWarn,message.c_str());
         ^
```
To avoid it, use something like this:
```
    LOG(logLevelWarn, "%s", message.c_str());
```
or even do this instead of constructing `message`:
```
    LOG(logLevelWarn, "Error on unregistering \"ca\" factory: %s", e.what());
```

PR #84, opened 2018-01-29
Revert "When caProvider is destroyed make sure all channels are cleared"

mrkraimer:
Reverts epics-base/pvAccessCPP#79


mrkraimer:
when exiting a crash occurs in CAChannel

mrkraimer:
I have a fix so I will prepare a new pull request  tomorrow

PR #86, opened 2018-01-31
Added error checks for invalid provider and NULL channel ptr.

bhill-slac:
pvinfo crashes w/ seg fault if you don't have a valid ca or pva provider.


mdavidsaver:
I made what I think is an functionally equivalent change with b05ec689d0192b925729b2446f7198d7e9303af2.  I didn't add any more error messages as there is already one for an invalid provider.  Also, ```DefaultChannelRequester::build()``` will print something when createChannel() returns NULL.

```
$ ./bin/linux-x86_64-debug/pvinfo -p invalid test
unknown provider name 'invalid', only 'pva' and 'ca' are supported
```

Of course, the fact that you saw this at all is a symptom of #87

PR #91, opened 2018-02-20
CAProvider: show which fields changed value

mrkraimer:
For get let the changed BitSet show which fields have changed value since the last get.
Similarly for monitor the changed BitSet shows which fields have changed since the last monitor event.

This is implemented via the pvCopy facility that is in pvData.
But an additional public method was requred in pvCopy.

### DIFF
--- a/src/ca/caChannel.cpp
+++ b/src/ca/caChannel.cpp
@@ -755,7 +755,8 @@ CAChannelGet::CAChannelGet(CAChannel::shared_pointer const & channel,
   :
     channel(channel),
     channelGetRequester(channelGetRequester),
-    pvRequest(pvRequest)
+    pvRequest(pvRequest),
+    firstTime(true)
 {
     if(DEBUG_LEVEL>0) {
         cout << "CAChannelGet::CAChannelGet() " << channel->getChannelName() << endl;
@@ -780,7 +781,9 @@ void CAChannelGet::activate()
     getType = getDBRType(pvRequest, channel->getNativeType());
     pvStructure = createPVStructure(channel, getType, pvRequest);
     bitSet = BitSetPtr(new BitSet(pvStructure->getStructure()->getNumberFields()));
-    bitSet->set(0);
+    pvCopy = PVCopy::create(
+          createPVStructure(channel, getType, pvRequest),
+          pvRequest,"");
     channel->addChannelGet(shared_from_this());
     if(channel->getConnectionState()==Channel::CONNECTED) {
          EXCEPTION_GUARD(getRequester->channelGetConnect(Status::Ok, shared_from_this(),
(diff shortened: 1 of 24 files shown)


src/ca/caChannel.cpp
```diff
@@ -793,14 +797,18 @@ void CAChannelGet::channelCreated(const Status& status,Channel::shared_pointer c
...
     if(!getRequester) return;
     chtype newType = getDBRType(pvRequest, channel->getNativeType());
     if(newType!=getType) {
         getType = getDBRType(pvRequest, channel->getNativeType());
         pvStructure = createPVStructure(channel, getType, pvRequest);
         bitSet = BitSetPtr(new BitSet(pvStructure->getStructure()->getNumberFields()));
-        bitSet->set(0);
+        pvCopy = PVCopy::create(
+          createPVStructure(channel, getType, pvRequest),
+          CreateRequest::create()->createRequest("field()"),
```
mdavidsaver: Is there more work to be done here?  Why is the provided ```pvRequest``` variable not passed to ```PVCopy::create()```?  I would naively assume that it should be.

src/ca/caChannel.cpp
```diff
@@ -1208,6 +1219,12 @@ void CAChannelGet::getDone(struct event_handler_args &args)
         {
             throw  std::runtime_error("CAChannelGet::getDone no copy func implemented");
         }
+        pvCopy->updateMasterSetBitSet(pvStructure,bitSet);
+        if(firstTime) {
+            bitSet->clear();
+            bitSet->set(0);
+            firstTime = false;
```
mdavidsaver: Unrelated to pvCopy, but the 'firstTime' member has a volatile qualifier which looks like it's meant to be used without a mutex.  This looks a bit iffy.  Then it is used here along with changes to ```bitSet``` which certainly isn't reentrant.  This doesn't seem right.  It seems to me that the volatile is unnecessary as it should be covered by any locking needed by ```bitSet``` (which may be none).

mrkraimer:
I agree that volatile is not necessary.
I think that no locking is needed since all use of firstTime  is in callbacks from ca

src/ca/caChannel.cpp
mrkraimer: The create code is:
PVCopyPtr PVCopy::create(
    PVStructurePtr const &pvMaster, 
    PVStructurePtr const &pvRequest, 
    string const & structureName)
{
    PVStructurePtr pvStructure(pvRequest);
    if(structureName.size()>0) {
        if(pvRequest->getStructure()->getNumberFields()>0) {
            pvStructure = pvRequest->getSubField<PVStructure>(structureName);
            if(!pvStructure) return PVCopyPtr();
        }
    } else if(pvStructure->getSubField<PVStructure>("field")) {
        pvStructure = pvRequest->getSubField<PVStructure>("field");
    }
    PVCopyPtr pvCopy(new PVCopy(pvMaster));
    bool result = pvCopy->init(pvStructure);
    if(!result) pvCopy.reset();
    return pvCopy;
}

The pvRequest structure is passed to pvCopy->init

mrkraimer:
Do to input from Michael Davidsaver,
I hove spent time looking at how ca provider  converts between DBD and pvData.
I now see a better way  to do the conversion that will not use pvCopy.
Thus I am canceling this pull request.

When I finish the new way for ca provider to convert between DBD and pvData,
I will issue a new pull request.

PR #107, opened 2018-06-26
server: pipeline=true windowing

mdavidsaver:
Implement flow control windows in the server itself, so that this applies to all Monitor,
even those which have no awareness of the flow control window.

Then the reportRemoteQueueStatus() callback is only needed to deliver the initial window update in order to keep track of the window size.

This is done by not calling Monitor::release() immediately when pipeline=true.  Rather, hold onto MonitorElement s until an ack is received.

Also addresses a race with clearing _channelMonitor vs. having send() queued at/after destroy().

### DIFF
--- a/src/server/pv/responseHandlers.h
+++ b/src/server/pv/responseHandlers.h
@@ -7,6 +7,8 @@
 #ifndef RESPONSEHANDLERS_H_
 #define RESPONSEHANDLERS_H_
 
+#include <list>
+
 #include <pv/timer.h>
 
 #include <pv/serverContext.h>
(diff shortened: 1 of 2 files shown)


mdavidsaver:
As discussed during the Hangout this morning, I'll leave this open for a day or two for comments.

Basic testing can be done with the 'spamme' example.

```sh
./examples/O.linux-x86_64/spamme
```

Then in another shell run

```sh
pvget -r 'field()record[pipeline=true]' -m spam1
```

Also test with QSRV (or any PVA server) to see that it now works with pipeline=true.

PR #108, opened 2018-06-27
a MonitorFIFO utility implementation of Monitior ABC

mdavidsaver:
At attempt to standardize the upstream interface of a Monitor FIFO, and hopefully be then only necessary Monitor implementation.  The intent is to capture all possible uses of Monitor, from basic fixed size FIFO, and including flow control (aka. pipeline=true).  Unlike previous implementations, it includes the notion of being disconnected, and handles type changes naturally.

Currently being used in P4P, will eventually be used in pva2pva.  This is a further generalization of BaseMonitor which is one of the common pieces which tie together P2P and QSRV.

http://mdavidsaver.github.io/pvAccessCPP/classepics_1_1pv_access_1_1_monitor_f_i_f_o.html#details

### DIFF
--- a/examples/spamme.cpp
+++ b/examples/spamme.cpp
@@ -32,6 +32,8 @@ typedef epicsGuard<epicsMutex> Guard;
 
 epicsEvent done;
 
+bool debug;
+
 #ifdef USE_SIGNAL
 void alldone(int num)
 {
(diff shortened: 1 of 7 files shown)


mdavidsaver:
Merged via manual rebase.

PR #111, opened 2018-06-30
Sharedpv

mdavidsaver:
A new server-side ChannelProvider API.  This is more task focused, and doesn't try for full generality (as is the case of ```pva/client.h```).  The use case being handled is the common one of a definite (though mutable) list of variables (PVs) to which some number of subscribers are interested (cf. ```StaticProvider```).  The core though is ```class SharedPV```.

```SharedPV``` has ```open(Field&)``` and ```close()``` methods to handle server side disconnect, and type change.  SharedPV is stateful, so a close() must be followed by an open() before clients see data.  An initial ```open(FIeld)``` is required.  Each call to ```open(Field&)``` may give a different type.

All subscribers are treated equal.  This API only exposes a minimum of information about the client list in general (including get, put, rpc, as well as monitor).  When that list becomes (non) empty.  This is handled in such as way that the type need not be determined until a client connects.  This allows for optimizations of server side processing (eg. db_event_enable()/db_event_disable() ).

A ```SharedPV``` is updated with ```post()``` which takes a PVStructure and bit mask.  The internal mutex is not exposed, so it is not possible to perform atomic read-modify-write operations on a SharedPV (it's not a record!).  SharedPV is better thought of as a cache to which server side code pushes updates to.

### DIFF
--- a/documentation/Doxyfile
+++ b/documentation/Doxyfile
@@ -768,6 +768,7 @@ WARN_LOGFILE           =
 INPUT                  = ../src/client/pv \
                          ../src/client/pva \
                          ../src/server/pv \
+                         ../src/server/pva \
                          ../src/rpcClient/pv \
                          ../src/rpcService/pv \
                          ../src/utils/pv \
--- a/documentation/examples.h
+++ b/documentation/examples.h
@@ -55,3 +55,9 @@ This example demonstrates a client which sets up a persistent Monitor operation.
 @include monitorme.cpp
 
 */
+/** @page examples_mailbox Server Mailbox Example
+This example creates a server with one or more pvas::SharedPV::buildMailbox()
+instances.
+
+@include mailbox.cpp
+*/
--- a/documentation/mainpage.h
+++ b/documentation/mainpage.h
@@ -8,6 +8,7 @@
 - @ref pvarelease_notes
 - API documentation
  - @ref pvac page
+ - @ref pvas page
  - @ref providers page
  - @ref pvtools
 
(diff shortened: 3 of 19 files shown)


mdavidsaver:
The dev. documentation has been updated

http://mdavidsaver.github.io/pvAccessCPP/group__pvas.html

several new examples are added

https://github.com/mdavidsaver/pvAccessCPP/tree/sharedpv/examples

as well as unit tests.

anjohnson:
> All subscribers are treated equal.

Will this become a problem when adding access controls (Access Security) later? We really should at least think about security when designing new APIs, even if we can't implement it properly throughout the stack yet.

mdavidsaver:
> Will this become a problem when adding access controls (Access Security) later?

Yes, though only for enforcing read restrictions.  Imo. both harder to implement, and less useful in controls applications.

> We really should at least think about security when designing new APIs

My plan (such as it is) includes delivering authentication information for Put and RPC through the ```pvac::Operation``` objects.

anjohnson:
> My plan (such as it is) includes delivering authentication information for Put and RPC through the `pvas::Operation` objects.

Would that exclude code using the older server API from being able to support auth*tion then? Just clarifying this so everyone understands the implications of this plan.

mdavidsaver:
> Would that exclude code using the older server API from being able to support auth*tion then?

No.  SharedPV et al. are convenience wrappers around the lower level APIs, and will remain so for the foreseeable future.  So exposing auth. info through the wrapper necessarily means exposing through the lower level APIs as well.

PR #120, opened 2018-08-01
fix issue 118; add put/get equal tests for pvScalar

mrkraimer:


mrkraimer:
Andrew,
Please merge this pull request

anjohnson:
Hi Marty,

With your changes to the test code I can add more calls to `checkClient()` for additional legal values, which is what I was looking for when I asked you to add more tests. I tried this and discovered the following:

1. DBRlongout tests work fine over the standard signed 32-bit decimal integer range. The checks below all pass and should be included in the test program:
```
        checkClient("DBRlongout", "0");
        checkClient("DBRlongout", "1");
        checkClient("DBRlongout", "-1");
        checkClient("DBRlongout", "32767");
        checkClient("DBRlongout", "32768");
        checkClient("DBRlongout", "-32768");
        checkClient("DBRlongout", "-32769");
        checkClient("DBRlongout", "2147483647");
        checkClient("DBRlongout", "-2147483648");
```

I can't use hexadecimal or octal numbers though, which surprised me slightly. The checks below all fail, because the test code converts `getValue` into a string which it then compares to the original instead of doing the comparison using the native type of the pvScalar:
```
        checkClient("DBRlongout", "0x10");
        checkClient("DBRlongout", "-0xf");
        checkClient("DBRlongout", "0xFEED");
        checkClient("DBRlongout", "07654");
```

I guess it isn't essential to be able to use hex or octal values in the tests, although it would make it easier to see at a glance that we're checking the right values if they could be expressed like this:
```
        checkClient("DBRlongout", "0x7FFF");
        checkClient("DBRlongout", "0x8000");
        checkClient("DBRlongout", "-0x8000");
        checkClient("DBRlongout", "-0x8001");
        checkClient("DBRlongout", "0x7FFFFFFF");
        checkClient("DBRlongout", "-0x80000000");
```

2. Another reason for doing value comparisons using the native type is because different OSs generate slightly different string outputs for floating-point values in exponential notation. The following tests pass on Linux, but the values with exponents could fail on some Windows versions because the OS defaults to providing 3 digit exponents instead of 2 (i.e. `1.5e-005`), and it may omit the + sign (I forget):
```
        checkClient("DBRdoubleout", "0");
        checkClient("DBRdoubleout", "1");
        checkClient("DBRdoubleout", "-1");
        checkClient("DBRdoubleout", "1.5");
        checkClient("DBRdoubleout", "-1.5");
        checkClient("DBRdoubleout", "1.5e-05");
        checkClient("DBRdoubleout", "1.5e+15");
        checkClient("DBRdoubleout", "-1.5e-05");
        checkClient("DBRdoubleout", "-1.5e+15");
        checkClient("DBRdoubleout", "1.5e-105");
```

Tests of DBRdoubleout should also include ±Infinity. I don't know why but these fail, I get ±1e+29 back instead of the value expected:
```
        checkClient("DBRdoubleout", "1.5e+51");
        checkClient("DBRdoubleout", "Inf");
        checkClient("DBRdoubleout", "-Inf");
```
I get ±1e+29 back from _any_ number larger than that, so there's one bug to be fixed.

Interestingly the value "nan" passes on Linux when compared as a string, but I don't think the C/C++ standard requires the string representation to always be spelled "nan" in all lower-case, which is what passes for me:
```
        checkClient("DBRdoubleout", "nan");
```
Checking for NaN as a value is more work since `NaN==<anything>` always returns False; you have to use `isnan()` on both sides. This is a pattern we've handled many times before though, for example epicsCalcTest.c in libCom/test. I'd be Okay with not checking NaN since we already know that both CA and PVA can pass it correctly.

3. I don't see any tests for scalars with type byte, signed & unsigned short or 32-bit float.

4. If you replaced the group "1 2 3" in the array tests with the complete set of test values used for the scalar equivalent you could check all those values for that array data type at once, for example:
```
        checkClient("DBRshortArray", "0 1 32767 -1 -32768");
```
You'll need to expand your waveform records beyond 5 elements or test the larger types in groups of 5 values.

5. I don't see any tests that are expected to fail; if you put the value 32768 to a short, does what you expect to happen actually happen?

6. This is probably a completely different issue, but does the CA provider support long strings (which look like char arrays to the CA client)? In 3.15 and up we have both lsi and lso record-types that you can experiment with, which are equivalent to a waveform with type CHAR.

I know doing this stuff can be rather boring, but by trying out everything I have today we know that there's something strange going on related to that 1e+29 value which needs more investigation.

- Andrew

mrkraimer:
Andrew,
Can you please merge now and then create a new issue about additional tests.
The reason is that I also want you to merge issue119.

I then want to create a branch issue117.
If issue119 is not merged first  that I suspect there may be a conflict because they involve the same file.

Marty

ralphlange:
I do not like this style of development.
Features and bug fixes should have tests that show they are working before getting merged.

anjohnson:
Marty, I resolve conflicts all the time when merging changes from Base-3.14 to Base-3.15 to Base-3.16 to Base-7.0, and in some cases that has required making fairly big changes while doing those up-merges. That's the nature of having to maintain multiple branches where we make bug fixes to code that has also been modified in later versions. That's another reason why self-tests that prove the bug has been fixed are so important, they should catch any mistakes made in the up-merging process.

In this case your original bug-fix only corrects issues with the DBRlongout tests, so Ralph's rule should be mostly satisfied if you add the checks in my first block of comments above. If you do that I will merge this.

mrkraimer:
issues 117, 118, and 119 are all related to ca provider.
I had been addressing them all in 
https://github.com/mrkraimer/pvAccessCPP

But when I asked for pull request Andrew requested that I address them separately.
So I did.
But these are related to the same set of code modules and conflicts may arise if not done carefully.

Also for this issue Andrew asked for LOTS of extra test code, which will take some time to implement.
I am not against the extra testing but I think it is important that 117, 118, and 119 be merged first so that anyone testing against master branches will not find the same problems.

Il also maintain say that requests for more test code can go on forever.
Just think about a complete set of test code for  pvData.
I will guess that the total amount of test code would be many many more lines of code than pvData itself.

Same for many of the other modules in both epics-base and the other V4 modules.

mrkraimer:
I have been busy on other things.
I will get back to this and add the checks in my first block of comments from Andrew's message.

PR #125, opened 2018-09-12
Changing default request string to 'field()'

es-sns121:
The default request string was changed from "field(value)" to "field()".

This allows ```pvget pv_name``` to work regardless if the pv has a value field or not. This is in line with the original pvget in which the entire pv structure would be dumped to stdout if the pv did not have a value field.

This unfortunately entails requesting the entire pv structure every time, but I could not figure out a way to modify the current form of pvget to check the structure and reissue a new request if the structure lacks a value field.

### DIFF
--- a/pvtoolsSrc/pvget.cpp
+++ b/pvtoolsSrc/pvget.cpp
@@ -42,7 +42,7 @@ namespace {
 
 bool debugFlag = false;
 
-string request("field(value)");
+string request("field()");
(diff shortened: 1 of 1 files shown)


mdavidsaver:
Apologies.  I missed the notification when this PR was opened (the 12th was a travel day for me).  The same change is made in #129, but this change is simple enough I'll merge it now.

PR #140, opened 2019-03-23
Env get config param

bhill-slac:
Adds a check for default EPICS_PVA_* env variables from EPICS_BASE/configure/CONFIG_ENVS
before using defaults from pvaConstants.h

See https://code.launchpad.net/~bhill/epics-base/+git/epics-base/+merge/358197

### DIFF
--- a/src/utils/configuration.cpp
+++ b/src/utils/configuration.cpp
@@ -119,9 +119,30 @@ void ConfigurationMap::addKeys(keys_t& names) const
         names.insert(it->first);
 }
 
+const ENV_PARAM * findEnvConfigParam( const char * envVarName )
+{
+    const ENV_PARAM **ppParam = env_param_list;
+
+    /* Find a match with one of the EPICS env vars */
+    while (*ppParam) {
+        if ( strcmp( envVarName, (*ppParam)->name ) == 0 ) {
+            return *ppParam;
+        }
+    ppParam++;
+    }
+
+    return 0;
+}
+
+
 bool ConfigurationEnviron::tryGetPropertyAsString(const std::string& name, std::string* val) const
 {
     const char *env = getenv(name.c_str());
+    if(!env) {
+        const ENV_PARAM *pParam = findEnvConfigParam( name.c_str() );
+        if ( pParam )
+            env = pParam->pdflt;
+    }
     if(!env || !*env)
         return false;
     if(val)


src/utils/configuration.cpp
```diff
@@ -122,8 +123,15 @@ void ConfigurationMap::addKeys(keys_t& names) const
 bool ConfigurationEnviron::tryGetPropertyAsString(const std::string& name, std::string* val) const
 {
     const char *env = getenv(name.c_str());
-    if(!env || !*env)
+    if(!env || !*env) {
+		const ENV_PARAM *pParam;
+		pParam = envFindConfigParam( name.c_str() );
+		if ( pParam && pParam->pdflt ) {
+			*val = pParam->pdflt;
```
mdavidsaver: My reason for passing `string* val` is to hint that it may be NULL.  cf. `Configuration::hasProperty()`.

src/utils/configuration.cpp
```diff
@@ -122,8 +123,15 @@ void ConfigurationMap::addKeys(keys_t& names) const
...
+			return true;
+		}
         return false;
+	}
```
mdavidsaver: The logic here can be slightly simplified as:

```c++
bool ConfigurationEnviron::tryGetPropertyAsString(const std::string& name, std::string* val) const
{
    const char *env = getenv(name.c_str());
    if(!env) {
        const ENV_PARAM *pParam = envFindConfigParam( name.c_str() );
        if(pParam)
            env = pParam->pdflt;
    }
    if(!env || !*env)
        return false;
    if(val)
        *val = env;
    return true;
}
```

src/utils/envFindConfigParam.cpp
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright - See the COPYRIGHT that is included with this distribution.
+ * pvAccessCPP is distributed subject to a Software License Agreement found
+ * in file LICENSE that is included with this distribution.
+ */
+
+#include <string.h>
+#include "envDefs.h"
+#include "envFindConfigParam.h"
+
+using namespace std;
```
mdavidsaver: Is this necessary?  No c++ headers are being included.

src/utils/configuration.cpp
bhill-slac: Yes, I should have included a test for NULL val ptr.   Hadn't seen hasProperty().

src/utils/envFindConfigParam.cpp
bhill-slac: No, I added that while looking into build error that turned out to be missing extern "C".

src/utils/configuration.cpp
bhill-slac: looks good.

anjohnson:
Core mtg: Move the envFindConfig routine and declaration into libcom/src/env in the LP merge.

bhill-slac:
Simplified patch and made independent of adding envFindConfigParam() to base.

mdavidsaver:
Ok.  The only additional change would be to mark `findEnvConfigParam()` static.  I'll do this after merging.

mdavidsaver:
Now that I've clicked merge, it occurs to me that the following (so far hypothetical) use case
would be broken.

```c++
    ... = ConfigurationBuilder()
            .push_env()
            .add("EPICS_PVAS_INTF_ADDR_LIST", "127.0.0.1")
            .push_map()
            .build();
```

This would have been a way to provide an application specific default.

I'm not bothered enough by this to do anything atm.

Issue #142, opened 2019-05-16
Echo doesn't, well, echo


kasemir:
#139 mentions that the C++ client may never actually send echo requests.
Working on a Java client that sends echo requests I noticed that the C++ server does reply to echo requests, but it always sends an empty reply.
According to the protocol, the echo should contain the payload of the echo request sent by the client, https://github.com/epics-base/pvAccessCPP/wiki/protocol#echo-0x02

I assume the issue is in https://github.com/epics-base/pvAccessCPP/blob/master/src/server/responseHandlers.cpp#L228 which ignores the received payload and always enqueues an empty echo reply:
```
void ServerEchoHandler::handleResponse(osiSockAddr* responseFrom,
                                       Transport::shared_pointer const & transport, int8 version, int8 command,
                                       size_t payloadSize, ByteBuffer* payloadBuffer)
{
    AbstractServerResponseHandler::handleResponse(responseFrom,
            transport, version, command, payloadSize, payloadBuffer);

    // send back
    TransportSender::shared_pointer echoReply(new EchoTransportSender(responseFrom));
    transport->enqueueSendRequest(echoReply);
}
```

Since the echo is requested and received via TCP, there is probably no need to check "is this the reply to _my_ echo request?". Still, a paranoid client might want to include a counter in the echo requests to check if the replies are matched.

mdavidsaver:
I agree that this is an omission which should be corrected.  Do we know if the Java server behaves similarly?

kasemir:
Yes, it's fully compatible.
Only difference is that the source code contains a commented-out section that tries to include the received payload.

https://github.com/epics-base/epicsCoreJava/blob/master/pvAccessJava/src/org/epics/pvaccess/server/impl/remote/handlers/EchoHandler.java

mdavidsaver:
Ah, I see.  Presumably because `payloadBuffer` is shared, and has been reused by the time the send() callback runs.

kasemir:
Assume the server is "fixed" and a client now depends on actually receiving the echoed messages. Such a client will fail with older servers, so now we need a new protocol version, ...
Based on the last core phone meeting, I don't see a way to overcome this.

mdavidsaver:
> so now we need a new protocol version, ...

Yes.

> I don't see a way to overcome this.

Increment the protocol version and see if anything breaks?  So far I've not come across any code which actually tests the protocol (minor) version.

kasemir:
[...]

PR #143, opened 2019-05-16
Deprecate protocol version 0

mdavidsaver:
Begin ignoring old version 0 peers.  Current version is 1 since 2014.  cf. #137

Also clean up debug printing of invalid messages.

PR #144, opened 2019-05-17
TCP Idle Timeout

mdavidsaver:
Prepare for adding TCP idle timeout.

Steps?

1. Fix server echo #142
2. Increment server protocol version
3. Enforce idle timeout If min(peer, self) version

1. Client send echo
2. Increment client protocol version
3. Enforce idle timeout If min(peer, self) version

### DIFF
--- a/TODO
+++ b/TODO
@@ -6,6 +6,5 @@
 
 - improve searching of channel with server address specified
 
-- void transportUnresponsive() is not implemented (also in Java)
 - complete authNZ (callback on right change)
 - request event on disconnect/destroy, etc.?
--- a/pvtoolsSrc/pvlist.cpp
+++ b/pvtoolsSrc/pvlist.cpp
@@ -120,6 +120,10 @@ bool processSearchResponse(osiSockAddr const & responseFrom, ByteBuffer & receiv
 
     // second byte version
     int8 version = receiveBuffer.getByte();
+    if(version == 0) {
+        // 0 -> 1 included incompatible changes
+        return false;
+    }
 
     // only data for UDP
     int8 flags = receiveBuffer.getByte();
--- a/pvtoolsSrc/pvlist.cpp
+++ b/pvtoolsSrc/pvlist.cpp
@@ -337,7 +337,7 @@ bool discoverServers(double timeOut)
     ByteBuffer sendBuffer(buffer, sizeof(buffer)/sizeof(char));
 
     sendBuffer.putByte(PVA_MAGIC);
-    sendBuffer.putByte(PVA_VERSION);
+    sendBuffer.putByte(PVA_CLIENT_PROTOCOL_REVISION);
     sendBuffer.putByte((EPICS_BYTE_ORDER == EPICS_ENDIAN_BIG) ? 0x80 : 0x00); // data + 7-bit endianess
     sendBuffer.putByte((int8_t)CMD_SEARCH);	// search
     sendBuffer.putInt(4+1+3+16+2+1+2);		// "zero" payload
--- a/src/pva/pv/pvaConstants.h
+++ b/src/pva/pv/pvaConstants.h
@@ -29,7 +29,7 @@ namespace pvAccess {
 const epics::pvData::int8 PVA_MAGIC = static_cast<epics::pvData::int8>(0xCA);
 
 const epics::pvData::int8 PVA_SERVER_PROTOCOL_REVISION = 2;
-const epics::pvData::int8 PVA_CLIENT_PROTOCOL_REVISION = 1;
+const epics::pvData::int8 PVA_CLIENT_PROTOCOL_REVISION = 2;
 
 /** PVA protocol revision (implemented by this library). */
 const epics::pvData::int8 PVA_PROTOCOL_REVISION EPICS_DEPRECATED = 1;
--- a/src/pva/pv/pvaConstants.h
+++ b/src/pva/pv/pvaConstants.h
@@ -7,6 +7,8 @@
 #ifndef PVACONSTANTS_H_
 #define PVACONSTANTS_H_
 
+#include <compilerDependencies.h>
+
 #ifdef epicsExportSharedSymbols
 #   define pvaConstantsepicsExportSharedSymbols
 #   undef epicsExportSharedSymbols
@@ -26,11 +28,14 @@ namespace pvAccess {
 /** PVA protocol magic number */
 const epics::pvData::int8 PVA_MAGIC = static_cast<epics::pvData::int8>(0xCA);
 
+const epics::pvData::int8 PVA_SERVER_PROTOCOL_REVISION = 1;
+const epics::pvData::int8 PVA_CLIENT_PROTOCOL_REVISION = 1;
+
 /** PVA protocol revision (implemented by this library). */
-const epics::pvData::int8 PVA_PROTOCOL_REVISION = 1;
+const epics::pvData::int8 PVA_PROTOCOL_REVISION EPICS_DEPRECATED = 1;
 
 /** PVA version signature used to report this implementation version in header. */
-const epics::pvData::int8 PVA_VERSION = PVA_PROTOCOL_REVISION;
+const epics::pvData::int8 PVA_VERSION EPICS_DEPRECATED = 1;
 
 /** Default PVA server port. */
 const epics::pvData::int32 PVA_SERVER_PORT = 5075;
--- a/src/pva/pv/pvaConstants.h
+++ b/src/pva/pv/pvaConstants.h
@@ -28,7 +28,7 @@ namespace pvAccess {
 /** PVA protocol magic number */
 const epics::pvData::int8 PVA_MAGIC = static_cast<epics::pvData::int8>(0xCA);
 
-const epics::pvData::int8 PVA_SERVER_PROTOCOL_REVISION = 1;
+const epics::pvData::int8 PVA_SERVER_PROTOCOL_REVISION = 2;
 const epics::pvData::int8 PVA_CLIENT_PROTOCOL_REVISION = 1;
 
 /** PVA protocol revision (implemented by this library). */
--- a/src/remote/abstractResponseHandler.cpp
+++ b/src/remote/abstractResponseHandler.cpp
@@ -4,8 +4,6 @@
  * in file LICENSE that is included with this distribution.
  */
 
-#include <sstream>
-
 #include <osiSock.h>
 
 #include <pv/byteBuffer.h>
@@ -14,9 +12,6 @@
 #define epicsExportSharedSymbols
 #include <pv/remote.h>
 #include <pv/hexDump.h>
-
-using std::ostringstream;
-using std::hex;
 
 using namespace epics::pvData;
 
@@ -44,13 +39,10 @@ void ResponseHandler::handleResponse(osiSockAddr* responseFrom,
         char ipAddrStr[48];
         ipAddrToDottedIP(&responseFrom->ia, ipAddrStr, sizeof(ipAddrStr));
 
-        ostringstream prologue;
-        prologue<<"Message [0x"<<hex<<(int)command<<", v0x"<<hex;
-        prologue<<(int)version<<"] received from "<<ipAddrStr<<" on "<<transport->getRemoteName();
-
-        hexDump(prologue.str(), _description,
-                (const int8*)payloadBuffer->getArray(),
-                payloadBuffer->getPosition(), static_cast<int>(payloadSize));
+        std::cerr<<"Message [0x"<<std::hex<<(int)command<<", v0x"<<std::hex
+                 <<int(version)<<"] received from "<<ipAddrStr<<" on "<<transport->getRemoteName()
+                 <<" : "<<_description<<"\n"
+                 <<HexDump(*payloadBuffer, payloadSize);
     }
 }
 }
--- a/src/remote/abstractResponseHandler.cpp
+++ b/src/remote/abstractResponseHandler.cpp
@@ -42,7 +42,7 @@ void ResponseHandler::handleResponse(osiSockAddr* responseFrom,
         std::cerr<<"Message [0x"<<std::hex<<(int)command<<", v0x"<<std::hex
                  <<int(version)<<"] received from "<<ipAddrStr<<" on "<<transport->getRemoteName()
                  <<" : "<<_description<<"\n"
-                 <<HexDump(*payloadBuffer, payloadSize);
+                 <<HexDump(*payloadBuffer, payloadSize).limit(0xffff);
     }
 }
 }
--- a/src/remote/beaconHandler.cpp
+++ b/src/remote/beaconHandler.cpp
@@ -37,8 +37,8 @@ void BeaconHandler::beaconNotify(osiSockAddr* /*from*/, int8 remoteTransportRevi
                                  PVFieldPtr /*data*/)
 {
     bool networkChanged = updateBeacon(remoteTransportRevision, timestamp, guid, sequentalID, changeCount);
-    if (networkChanged)
-        changedTransport();
+    // TODO: reduce search timers
+    (void)networkChanged;
 }
 
 bool BeaconHandler::updateBeacon(int8 /*remoteTransportRevision*/, TimeStamp* /*timestamp*/,
@@ -81,19 +81,6 @@ bool BeaconHandler::updateBeacon(int8 /*remoteTransportRevision*/, TimeStamp* /*
     }
 
     return false;
-}
-
-void BeaconHandler::changedTransport()
-{
-    TransportRegistry::transportVector_t transports;
-    _context.lock()->getTransportRegistry()->toArray(transports, &_responseFrom);
-
-    // notify all
-    for (TransportRegistry::transportVector_t::iterator iter(transports.begin()), end(transports.end());
-         iter != end; iter++)
-    {
-        (*iter)->changedTransport();
-    }
 }
 
 }
--- a/src/remote/blockingUDPTransport.cpp
+++ b/src/remote/blockingUDPTransport.cpp
@@ -9,6 +9,8 @@
 #include <ws2tcpip.h>
 #endif
 
+#include <sstream>
+
 #include <sys/types.h>
 #include <cstdio>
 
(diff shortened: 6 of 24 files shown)


mdavidsaver:
So far this PR (maybe) actually echos.  Increments protocol version, both server and (erroneously) client.

@kasemir  Is this enough for you to test the java client against?

kasemir:
Hmmm.
I think I applied those changes to my copy of EPICS base 7.0.2.2 modules/pvAccess.

Good:
The old PV Access Java client that uses PVA_PROTOCOL_REVISION 1 still works.

My new PVA Java client also works like before when I send PVA_PROTOCOL_REVISION 2 but with an 'empty' echo request.
When I add a payload to my echo request, however, I seem to get something different echoed back.

From my side it looks like this:
```
2019-05-17 08:56:18 FINE [org.epics.pva] Sending ECHO request
2019-05-17 08:56:18 FINER [org.epics.pva] TCP send
0000 - CA 02 00 02 04 00 00 00 65 63 68 6F             - ........echo    
2019-05-17 08:56:18 FINER [org.epics.pva] TCP receiver /127.0.0.1:5075: 12 bytes
2019-05-17 08:56:18 FINER [org.epics.pva] TCP receiver /127.0.0.1:5075:
0000 - CA 02 40 02 04 00 00 00 88 07 00 10             - ..@.........    
2019-05-17 08:56:18 FINE [org.epics.pva] Received ECHO:
0000 - 88 07 00 10                       
```
So I do get 4 bytes echoed back, but the content is not what I sent.

When I run softIocPVA with EPICS_PVA_DEBUG=100, it proves that my echo request indeed contains 'echo':
```
Message [0x2, v0x2] received from 127.0.0.1:52944 on 127.0.0.1:52944
Hexdump [Echo request] size = 4
65 63 68 6F                                          echo 
```

Need to check if I mis-applied the patch, or if there's indeed a bug that causes the server to send the wrong content.

kasemir:
Flabbergasted.
Placed a printf in EchoTransportSender::send:
```
   virtual void send(epics::pvData::ByteBuffer* buffer, TransportSendControl* control) OVERRIDE FINAL {
...
        buffer->putArray<char>(&toEcho[0], toEcho.size());
        if (toEcho.size()==4)
            printf("Send ECHO %c%c%c%c\n", toEcho[0], toEcho[1],toEcho[2],toEcho[3]);
```
It thinks it's echoing the correct payload content.
But when I check with wireshark, it's not.

According to wireshark, I send an echo request with payload 'echo':
sent

The reply contains a payload of 4 bytes, but it's not the 'echo' that it thinks it returns:
receive

mdavidsaver:
I didn't test this last night as I wasn't sure how.  I'll take a look at a little while and see if I can see what is going on.

If you want to look further, I'd suggest looking at the contents of the ByteBuffer and `ByteBuffer::getPosition()` and getLimit().  The actual send() happens in `BlockingTCPTransportCodec::write()` and uses the equivalent of:

```c++
        int bytesSent = ::send(_channel,
                               &src->getArray()[src->getPosition()],
                               src->getRemaining(), 0);
```

The position and remaining here should by changed by `ByteBuffer::flip()` in `AbstractCodec::flushSendBuffer()` so that the header is included.  The whole process should I think be happening within `AbstractCodec::processSendQueue()`.

kasemir:
When adding the 'toEcho' data, the buffer->getPosition() moves from 8 to 12, which seems correct: Plain PVA header size is 8, and we add the 4 'echo' bytes.
Tried running softIocPVA in gdb with `break EchoTransportSender::send` but can't get it to stop there.

kasemir:
The bug is in the `EchoTransportSender` constructor.
It copies the correct number of bytes, but not the actual bytes:
```
 EchoTransportSender(osiSockAddr* echoFrom, size_t payloadSize, epics::pvData::ByteBuffer& payloadBuffer) {
        memcpy(&_echoFrom, echoFrom, sizeof(osiSockAddr));
        toEcho.resize(payloadSize);
        memcpy(&toEcho[0], payloadBuffer.getArray(), payloadSize);
        if (payloadSize==4)
            printf("Got ECHO %c%c%c%c\n", toEcho[0], toEcho[1],toEcho[2],toEcho[3]);
    }
```

When I send an echo request with 4 bytes, the "Got ECHO ...." printout shows nonsense for the 4 bytes, not the 4 bytes that are in the payload.
I assume the `payloadBuffer.getArray()` is wrong. Needs a '+8' or something relative to 'getPosition'.

mdavidsaver:
> I assume the `payloadBuffer.getArray()`

I should have used `void ByteBuffer::getArray(T* values, std::size_t count)`, which takes `_position` into account.  The fact that some methods of ByteBuffer reference the current position and some don't has caused me confusion in the past as well...

kasemir:
From byteBuffer.h:
```
  // TODO remove
    EPICS_ALWAYS_INLINE const char* getArray() const
```

Fixed it like this:
```
EchoTransportSender(osiSockAddr* echoFrom, size_t payloadSize, epics::pvData::ByteBuffer& payloadBuffer)
{
    memcpy(&_echoFrom, echoFrom, sizeof(osiSockAddr));
    toEcho.resize(payloadSize);
    char *fu = &toEcho[0];
    payloadBuffer.getArray(fu, payloadSize);
}
```

Now it copies the payload and echoes it back, jay!

kasemir:
I've updated https://github.com/epics-base/pvAccessCPP/wiki/protocol:
Section about protocol versions, note on Echo command that only V 2 supports the payload.

mdavidsaver:
Bonus points...

```c++
    /**
     * Get the raw buffer data.
     * @return the raw buffer data.
     */
    inline const char* getBuffer() const
    {
        return _buffer;
    }
...
    // TODO remove
    EPICS_ALWAYS_INLINE const char* getArray() const
    {
        return _buffer;
    }
```

kasemir:
Alright. Still, `payloadBuffer.getArray(&toEcho[0], payloadSize);` would work, so update the PR and merge?

mdavidsaver:
Merging the fix for echo, sure.  I can't merge the version increment yet.  Is one useful to you without the other?

I can't increment the c++ client version until it is also sending CMD_ECHO (or the server should timeout).  Right now this code has no idea of a separate client and server version.  So version handling needs to be split and only the server incremented, and/or the client needs to send Echo.

I think both should be done eventually.  I'd be surprised if this were the only time we want to increment only one.  I'll take a look and see which would be easier to do first.

mdavidsaver:
Significant rework to (hopefully) do both sides of timeout.  Both client and server now identify as v2.

Timeout in this case is using `SO_RCVTIMEO`, and expires if no message (not just `CMD_ECHO`) is received within the timeout period.  This period being defined by `$EPICS_PVA_CONN_TMO`, and defaulting to 30 seconds.

The client sends beacons at half the timeout period (default 15 seconds).

This PR also includes the changes from #143 to ignore (incompatible) version 0 peers.  This isn't strictly speaking required, but the same testing is needed for both.

mdavidsaver:
Also worth noting that I'm initially enabling the timeout on every connection, then disabling it if the peer identifies as v1.  This should timeout peers which aren't actually speaking PVA.  This should be safe as actual PVA client should immediately begin the auth. handshake.  Also, the default 30 second idle timeout is longer than the current (unconfigurable) auth. timeout of 6 seconds.

mdavidsaver:
Also, I really do need Windows testing on this one!  `SO_RCVTIMEO` behaves (at least) slightly differently on windows.  What I have works with Linux+WINE, but that's not the important case.

mdavidsaver:
@kasemir As mentioned last week, could you retest against your work on the Java client.

@MarkRivers @anjohnson @ralphlange I could use some help testing this change on non-Linux targets.  Specifically the `#ifdef _WIN32` case.

The most important (regression) test is to run `pvmonitor` against a PV which isn't changing and ensure it does __not__ timeout.  This should be repeated for 3 combinations of client+server: new+old, old+new, new+new.

The next test is to suspend, or otherwise interrupt network communication, between client and server and see that both sides do timeout.  I've tried this on Linux, but I don't know of an equivalent to SIGSTOP on windows.

kasemir:
I fetched base 7.0.2.2 and then replaced modules/pvAccess with `git clone https://github.com/mdavidsaver/pvAccessCPP.git  && mv pvAccessCPP pvAccess`.

The result seems to include not only the echo/timeout fix but also some new auth requirements.
When I try to connect to softIocPVA, it complains "an exception caught while in receiveThread at ../../src/remote/codec.cpp:1137: 127.0.0.1:38222 failing attempt to select non-existant auth. plugin".

I assume I need to reply that I want to use e.g. "anonymous" authorization?

kasemir:
The protocol description mentions 'auth', but no detail on how to actually use that.
remote/pv/security.h includes source comments for dot diagrams, is that anywhere available as e.g. doxygen web pages?

kasemir:
Muddled further by sending "anonymous" authorization in the connection validation phase.

But echo is missing the content.
As a client, when I send an echo request after a channel is idle for ~15 seconds, I get empty content back.

I send the text 'echo' in the message body,
```
0000 - CA 02 00 02 04 00 00 00 65 63 68 6F             - ........echo
```
but the reply is empty:
```
0000 - CA 02 40 02 04 00 00 00 00 00 00 00             - ..@.........    
```

kasemir:
Fix for pvAccess/src/server/pv/responseHandlers.h:

```
EchoTransportSender(osiSockAddr* echoFrom, size_t payloadSize, epics::pvData::ByteBuffer& payloadBuffer) {
    memcpy(&_echoFrom, echoFrom, sizeof(osiSockAddr));
    toEcho.resize(payloadSize);
    // BAD:
    // memcpy(&toEcho[0], payloadBuffer.getBuffer(), payloadSize);
    // Good:
    payloadBuffer.getArray(&toEcho[0], payloadSize);
}
```

mdavidsaver:
> The result seems to include not only the echo/timeout

This PR is for the [idletimeout](https://github.com/mdavidsaver/pvAccessCPP/tree/idletimeout) branch.

kasemir:
Sorry, I started at "Steps? 1) Fix server echo #142" which I found broken.
With that fixed, my client gets 'echo' replies when it tests an idle connection every ~15 seconds.
What else should I try?

kasemir:
Yes, I am on the idletimeout branch, forgot to mention that in the git clone example.
But I only got your pvAccessCPP, the rest is original base 7.0.2.2.

mdavidsaver:
> remote/pv/security.h includes source comments for dot diagrams

Now at: http://mdavidsaver.github.io/pvAccessCPP/pva_security.html

kasemir:
PVA Sec. diagram is actually helpful, thanks!

gregoryraymondwhite:
Brilliant!

mdavidsaver:
> Fix for pvAccess/src/server/pv/responseHandlers.h:

I'm sure I applied this before, but ... apparently not.  Probably just me going crazy(er).  5abfc10a828311a83dee7da43ff6a6ccd0130066 is now pushed.

mdavidsaver:
> The result seems to include not only the echo/timeout fix but also some new auth requirements.
> When I try to connect to softIocPVA, it complains "an exception caught while in receiveThread at ../../src/remote/codec.cpp:1137: 127.0.0.1:38222 failing attempt to select non-existant auth. plugin".

This is from epics-base/pvAccessCPP#136. Which does add some validation.  It is no longer permitted to send an empty plugin name string.  The older clients I tested all selected the "ca" plugin.

https://github.com/epics-base/pvAccessCPP/blob/e66893bb38ed47ec6255cf6dd8911f4549b2d5ef/src/remote/codec.cpp#L1634-L1639

Compatible clients can either select "anonymous" with no data, or "ca" with a Structure including at least the "host" and "user" fields.

https://github.com/epics-base/pvAccessCPP/blob/e66893bb38ed47ec6255cf6dd8911f4549b2d5ef/src/remote/security.cpp#L24-L29

You'll know it is working if you don't see this error, and if `pvasr 5` (with QSRV) shows something like the following for a connected client.  The lines after `aux. auth:` are verbatim from the client.

```
epics> pvasr 5
Clients:
  tcp://192.168.210.1:35152  ver=2 1 channels user: ca/mdavidsaver groups:... aux. auth.:
    structure 
        string user mdavidsaver
        string host md-laptop

  cnt   : RW mdavidsaver@192.168.210.1, role/....
```

mdavidsaver:
> What else should I try?

The three tests I performed where as follows.  Run client and server with a PV which doesn't change

1. Wait to see that no timeout occurs (at least 30 seconds).
1. Suspend (SIGSTOP) the client.  See that the server times out after 30 seconds.  `pvasr 1` should show no clients and `refshow 1` should show zero `BlockingTCPTransportCodec` instances.
1. Suspend the server.  See that the client times out and beings searching again.

At least the first test should be repeated with an older v1 server as well, to check the compatibility logic.

kasemir:
Thanks for the auth detail, I'll look at that later. For now I'm sending 'anonymous', and I think the timeout works OK.

1) No timeout for long running monitor of static channel
Started softIocPVA with a database that contains a non-changing ai record. Monitor that with my client. Every ~15 second, my client sends an echo request because the TCP connection is otherwise idle. I now get the echo replies with correct payload. This ran fine over night, echo request initiated by client every ~15 seconds, neither server nor client disconnect, 

2) Server times out on dead client
Next I disabled the echo request on the client.
Client connects fine, gets the initial update, then does NOT send an echo request. As a consequence, the server closes the TCP connection after ~30 seconds. The server doesn't send an echo request, it simply closes the connection. Client then notices the closed TCP connection, starts searching again, server replies, they connect, repeats.
Not sure if the server should do anything other than close the connection on an idle client. First expected server to send an echo request to the client, but this will create a situation where every 15 seconds both the server will request echo from client and vice versa. Probably best for client to request echo, and server simply to close TCP connection.

3) Client times out on dead server
Next I sent SIGSTOP to the server. In that case, my client sends echo requests, doesn't get a reply, closes the connection, starts searching again. When I then continue the server, it replies to the search requests, they connect, repeat.

kasemir:
Regarding "ca" instead of "anonymous":

The 'TODO' sections of the protocol and the diagram suggest that the client only mentions auth="ca" in the connectionValidationResponse, and the server will then prompt for detail like user and host in follow-up authNZRequest/authNZResponse messages.

Instead, it looks like the server expects the user and host to follow in the connectionValidationResponse, and if that's not there, it crashes:
```
Program terminated with signal 11, Segmentation fault.
(gdb) bt
#0  epics::pvData::PVStructure::getSubFieldImpl (this=<optimized out>, name=name@entry=0x7fe824b9bef3 "user", throws=throws@entry=true) at ../../src/factory/PVStructure.cpp:150
#1  0x00007fe824b14b81 in getSubFieldT<epics::pvData::PVString, char const*> (a=0x7fe824b9bef3 "user", this=<optimized out>, this=<optimized out>, this=<optimized out>, 
    this=<optimized out>) at /home/controls/base-7.0.2.2/include/pv/pvData.h:803
#2  (anonymous namespace)::CAPlugin::createSession (this=0x7fe7b4000ce0, peer=std::tr1::shared_ptr (count 1, weak 0) 0x7fe7b00009f0, 
    control=std::tr1::shared_ptr (count 10, weak 1) 0x7fe7bc0009e0, data=std::tr1::shared_ptr (empty) 0x0) at ../../src/remote/security.cpp:99
#3  0x00007fe824b0b87c in epics::pvAccess::detail::BlockingServerTCPTransportCodec::authNZInitialize (this=this@entry=0x7fe7bc0008c0, securityPluginName="ca", 
    data=std::tr1::shared_ptr (empty) 0x0) at ../../src/remote/codec.cpp:1676
#4  0x00007fe824b42bac in epics::pvAccess::ServerConnectionValidationHandler::handleResponse (this=<optimized out>, responseFrom=<optimized out>, transport=..., 
    version=<optimized out>, command=<optimized out>, payloadSize=<optimized out>, payloadBuffer=0x7fe7bc000908) at ../../src/server/responseHandlers.cpp:207
#5  0x00007fe824b40642 in epics::pvAccess::ServerResponseHandler::handleResponse (this=<optimized out>, responseFrom=<optimized out>, transport=..., version=<optimized out>, 
    command=<optimized out>, payloadSize=<optimized out>, payloadBuffer=0x7fe7bc000908) at ../../src/server/responseHandlers.cpp:170
#6  0x00007fe824b11b65 in epics::pvAccess::detail::BlockingTCPTransportCodec::processApplicationMessage (this=0x7fe7bc0008c0) at ../../src/remote/pv/codec.h:362
```

Suggest that server checks payloadSize. If the expected user/host detail is not provided, issue error, but don't crash.

kasemir:
Specific bug seems to be this:
In src/server/responseHandlers.cpp around line 192, `ServerConnectionValidationHandler::handleResponse` leaves `data` empty when no payload content follow after `authNZ == "ca"`.
The CA plugin is nevertheless found and called. So CAPlugin could check for data being null or empty and refuse to authorize.

kasemir:
Got it to work with "ca":
```
epics> pvasr 5
Clients:
  tcp://127.0.0.1:43432  ver=2 1 channels user: ca/Fred aux. auth.:
    structure 
        string user Fred
        string host my_host
```

In the protocol description, https://github.com/epics-base/pvAccessCPP/wiki/protocol#connection-validation-0x01, I guess the TODO authNZRequest, authNZResponse can be deleted?
Instead, connectionValidationResponse changes to this:

```
struct connectionValidationResponse {
    int clientReceiveBufferSize;
    short clientIntrospectionRegistryMaxSize;
    short connectionQos;
    string authNZ;         // selected authNZ plugin;
    // Optional, content depends on authNZ
    FieldDesc dataIF;  
    PVField data;
};

Supported authNZ modes are "anonymous" and "ca".
For "ca", a structure with string elements "user" and "host" needs to follow.
```

mdavidsaver:
> So CAPlugin could check for data being null or empty and refuse to authorize.

Something quick which should at least not crash 05bf0fc453a9ae4de9007fbe7ba7da128fef4538

kasemir:
I've tried the "CAPlugin handle peer NULL data" update.
Looks good:
If client sends "ca" but then no user/host, the server closes the connection, no more crash.
If client does send the complete info, it can read.

mdavidsaver:
Ok.  Still no Windows testing.  I going to go ahead and merge.   The benefits of having a timeout are clear.  Any win32 specific issues will be addressed as they are discovered.

PR #151, opened 2019-07-22
TX/RX Stats

mdavidsaver:
Keep and expose statistics on number of bytes sent/received per socket, and per operation.

Will be used by the PVA gateway to track client bandwidth utilization.

The new `struct NetStats` is not placed in the public portion of the ChannelProvider class hierarchy as it will not be implemented by many providers (probably only the PVA client and server).  It does not fit with the symmetric idea of ChannelProvider.  eg. It is implemented by PVA client for `class ChannelGet`, but for `class ChannelGetRequester` by PVA server.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,4 +28,3 @@ env:
   - BRBASE=7.0 RTEMS=4.9 TEST=NO
   - BRBASE=3.16
   - BRBASE=3.15
-  - BRBASE=3.14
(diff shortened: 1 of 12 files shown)


mdavidsaver:
This PR would be entirely unremarkable except that it adds unconditional use of `epicsAtomic.h` and thus requires Base >=3.15.

If/when this PR is merged, pvAccessCPP will __no longer build against Base 3.14__!

mdavidsaver:
Silence is taken as acceptance.  I'll be merging and removing the 3.14 compatibility `#ifdef`s from this module, and then pvDataCPP as well.

PR #154, opened 2019-09-26
Tweak pvlist output to make it easier to cut/paste

bhill-slac:
Prior syntax had no spaces around ip_addr:port which results in tcp@[ and ] getting 
picked up w/ mouse based word selection.
Old syntax:
GUID 0x0863D45C00000000B8AAFA06 version 1: tcp@[172.21.68.76:47626]
New syntax:
GUID 0x0863D45C00000000B8AAFA06 version 1: tcp@[ 172.21.68.76:47626 ]


mdavidsaver:
Fair enough.

PR #162, opened 2020-04-16
testCaProvider on Windows with Base-3.15

anjohnson:
This should let testCaProvider start its external softIoc properly on Windows when built with a Base version too old to provide dbUnitTest.

I haven't actually tested this on Windows myself, but it does work on Darwin and I'm using OS-Independent APIs so I'm reasonably confident that it should work.


mdavidsaver:
Seems like this changed had the desired effect.  https://ci.appveyor.com/project/mdavidsaver/pvaccesscpp

Issue #163, opened 2020-05-04
EPICS base 7.0.3.1, deadlock


SkyToGround:
Calling the destructor on more than one instance of a pvac::ClientProvider which has been constructed with a provider name of "ca" (channel access) causes (what appears to be) a deadlock. Using "pva" (pvAccess) does not appear to trigger this deadlock. I have attached a screenshot of where the deadlock appears to be located (from pausing the application when running it under a debugger). The code that causes the triggers the deadlock can be found here: https://github.com/ess-dmsc/forward-epics-to-kafka/blob/a75fab2a7343906c147722825a258332fc2126e7/src/EpicsClient/EpicsClientMonitorImpl.h

In this piece of code, the version of EPICS base used is 7.0.3.1. I believe I had a the same issue when testing with  earlier versions of EPICS7 a few months back as well.
<img width="1904" alt="deadlock_screenshot" src="https://user-images.githubusercontent.com/58835/80964815-cd68b100-8e11-11ea-9193-4fbda163b700.png">

I have also attached screenshots of stack traces of other threads executing EPICS code.
<img width="740" alt="thread_8" src="https://user-images.githubusercontent.com/58835/80965309-9515a280-8e12-11ea-960c-5b4269b5adb7.png">
<img width="740" alt="thread_3" src="https://user-images.githubusercontent.com/58835/80965312-9777fc80-8e12-11ea-8963-498aeca123fc.png">
<img width="740" alt="thread_2" src="https://user-images.githubusercontent.com/58835/80965314-98a92980-8e12-11ea-8f82-ca209779b710.png">

mdavidsaver:
The main thread is waiting for the first of four caProvider worker threads to exit.  Can you say if this thread (in `PutDoneThread::run()`) is still running?  If so, please take a stack trace.  If not, can you say if there was a prior error message about eg. an unhandled exception?

https://github.com/epics-base/pvAccessCPP/blob/0897836520457138daec72eb09ffad8ec230a390/src/ca/caProvider.cpp#L80-L83

Also, you're emails mentioned that you had made a workaround.  The details of what you had to change to do this might also give some hints.

mdavidsaver:
Also, since trace for the main thread has been cropped.  Is the `main()` function still in progress, or has it returned?

SkyToGround:
* Yes, it is the main thread/function that has deadlocked (well) before reaching `return 0`.
[...]

Issue #165, opened 2020-06-10
testCaProvider deadlock


mdavidsaver:
I happened to time it right to catch a hung appveyor run for Base 7.0.  I found `testCaProvider.exe` still running after ~45 min.  I first tried killing caRepeater, with apparent effect.  The CI run completed shortly after I then killed the test.  The last step caused some new output to appear in the log.

https://ci.appveyor.com/project/mdavidsaver/epics-base/build/job/rf9ojmbx2qmj0rpn

```
perl -CSD testCaProvider.t -tap > testCaProvider.tap
epicsThread::~epicsThread(): "winadc" blocking for thread "channelConnectThread" to exit
was epicsThread object destroyed before thread exit ?
epicsThread::~epicsThread(): "winadc" blocking for thread "channelConnectThread" to exit
was epicsThread object destroyed before thread exit ?
...
```

mdavidsaver:
I don't think this is exactly the same as #163, since I think only one provider instance is created.  It looks like an issue with the same code though.

anjohnson:
Thanks.

From that error message I suspect I know what the issue may be here, caused by deleting an epicsThread object in a different thread after the child has already called exitWait() or returned. I already have a fix and tests for the 3.15 branch but it needs some work to properly merge up due to conflicts with the epicsThreadJoin() changes in 7.0. I will bump that upwards in priority, and also work on #163

mdavidsaver:
@anjohnson Is your fix imminent?  If not, I'm going to temporarily disable testCaProvider from running on windows.  Having eg. 3 appveyor jobs timeout during [one run](https://ci.appveyor.com/project/mdavidsaver/epics-base/builds/33640567) is annoying.

```diff
diff --git a/testCa/Makefile b/testCa/Makefile
index 1b1ded0f..a575235a 100644
--- a/testCa/Makefile
+++ b/testCa/Makefile
@@ -11,7 +11,9 @@ PROD_SYS_LIBS_WIN32 += netapi32 ws2_32
 
 TESTPROD_HOST += testCaProvider
 testCaProvider_SRCS += testCaProvider.cpp
+ifneq (,$(filter win%,$(T_A)))
 TESTS += testCaProvider
+endif
 ifdef BASE_3_16
   testCaProvider_SRCS += testIoc_registerRecordDeviceDriver.cpp
   REGRDDFLAGS = -l
```

anjohnson:
[...]

Issue #83, opened 2018-01-27
CA provider's threadAttach() doesn't disconnect an already-attached context


anjohnson:
If a program creates its own CA context and also happens to use the PVA CA provider, the threadAttach() method's call to
```
     ca_attach_context(current_context);
```
will fail silently, and subsequent CA calls inside the provider will fail. Fixing this will probably require some re-engineering of how the context is set to implement an RAII object that saves the CA context of the caller and temporarily instantiates our own, then restores theirs on exit if it was non-zero (similar to a mutex guard object).

Low priority since this isn't likely to be a common thing to want to do.

anjohnson:
Assume closed by #167

PR #167, opened 2020-06-29
Fixing #165

anjohnson:
Incomplete, hoping this PR will trigger an Appveyor build...


src/ca/caContext.h
```diff
@@ -0,0 +1,51 @@
...
+class Attach
+{
+public:
+    Attach(const CAContextPtr & to) :
+        context(to), saved_context(to->attach()) {}
+    ~Attach() {
+        context->detach(saved_context);
+    }
+private:
+    CAContextPtr context;
```
anjohnson, line 45: Make this a CAContext &

src/ca/caContext.cpp
```diff
@@ -0,0 +1,66 @@
...
+#define epicsExportSharedSymbols
+#include "caContext.h"
+
+namespace epics {
+namespace pvAccess {
+namespace ca {
+
+CAContext::CAContext()
+{
+    int result = ca_context_create(ca_enable_preemptive_callback);
```
anjohnson, line 19: Check if there's already a context, throw logic-error?

src/ca/notifierConveyor.cpp
```diff
@@ -0,0 +1,94 @@
...
+        {
+            NotificationWPtr notificationWPtr(workQueue.front());
+            workQueue.pop();
+            work = !workQueue.empty();
+            NotificationPtr notification(notificationWPtr.lock());
+            if (notification) {
+                notification->queued = false;
+                NotifierClientPtr client(notification->client.lock());
+                if (client) {
+                    the.unlock();
```
anjohnson, line 74: Think about client's destructor trying to un-queue itself, lock held or not?

src/ca/caContext.h
anjohnson: This works, @mdavidsaver I assume this is what you meant:
```patch
diff --git a/src/ca/caContext.h b/src/ca/caContext.h
index 620cb20f..6830c847 100644
--- a/src/ca/caContext.h
+++ b/src/ca/caContext.h
@@ -37,12 +37,12 @@ class Attach
 {
 public:
     Attach(const CAContextPtr & to) :
-        context(to), saved_context(to->attach()) {}
+        context(*to), saved_context(context.attach()) {}
     ~Attach() {
-        context->detach(saved_context);
+        context.detach(saved_context);
     }
 private:
-    CAContextPtr context;
+    CAContext & context;
     ca_client_context* saved_context;
 };
 
```

mdavidsaver: Yes.

Also, as a matter of habit, I start by marking any single argument constructor with the [`explicit`](https://en.cppreference.com/w/cpp/language/explicit) specifier to inhibit some implicit casts.  I rarely intend to allow sure of such casts, and this is a way of preventing surprise.

anjohnson: `explicit` added to the constructor, thanks.

src/ca/caContext.cpp
anjohnson: There seems no need to throw here if the thread is already attached, creating a new CAContext doesn't need to leave us attached to it anyway, so if the thread was already attached to one I restore it before returning. There was a bug in the destructor though.

src/ca/notifierConveyor.cpp
anjohnson: There is no API for a client to unqueue itself, only the conveyor thread can remove items from the queue.

Looking at `NotifierConveyor::run()` the mutex is taken at the top of the outer loop for access to the conveyor's data members, and unlocked before calling a `NotifierClient::notifyClient()` method. After that returns if there were more items on the queue (having read that into `work` before we unlocked) we must take the lock again before we loop back to get the next item. If the queue was empty we will fall out of the inner loop without taking the lock again. Thus when the `NotifierClientPtr client` goes out of scope we have two cases to consider, since the lock may be held or not depending whether `work` was set.

If `client` holds the last reference to a `NotifierClient` object the `shared_ptr` destructor might result in other objects being deleted, the interesting ones here being `Notification` and/or `NotifierConveyor`.

Our strong reference `notification` will slightly out-last the `client` object, but it's destruction has no knock-on effects.

The `CAChannelProvider` contains the two `NotifierConveyor` data members. Are there any restrictions on deleting a ChannelProvider from one of its Requester callbacks?

If this `NotifierConveyor` object gets deleted by the `NotifierClient` destructor, the conveyor's destructor will attempt to take the mutex. It doesn't matter whether we already own it or not though because the recursive nature of the mutex permits the destructor to acquire the lock immediately or after a short delay if some other thread is queueing a notification. The destructor sets `halt`, releases the lock and signals the Event, then calls `thread->exitWait()`. The semantics of that `epicsThread` method ensure that since it is being called from inside the thread it will return immediately, permitting the destructor to exit. At this point the conveyor object data members may no longer be accessed, which is a problem if `work` is true or `stopping` hasn't been set from `halt` by then.

Changing the `run()` method to solve that...

mdavidsaver:
@anjohnson fyi. If you rebase this branch you should (for the 7.0 builds) have automatic analysis of any tests which crash.

src/ca/notifierConveyor.cpp
mdavidsaver: I think I may not have made myself clear.  Of course you can, with sufficient effort, convince yourself that a particular scenario is not a concern.

The point I was trying to make, and which may have been lost in the weeds, is that proper attention to _local_ shared_ptr scope can avoid the need to perform a _global_ analysis.  This is a matter of better practice (I won't claim it's best) which I've come to after wrestling with this code base for the past ~3 years.  As with locking (I guess I failed to draw that analogy) I find it easier to work with rules which I can apply to pieces of code in isolation.

To return to the specific, I would write the skeleton of a worker thread of this sort as:

```c++
struct myworker {
    epicsMutex lock;
    epicsEvent wakeup;
    std::list<std::weak_ptr<...>> work;
    bool running;

    void run() {
        epicsGuard<epicsMutex> G(lock);
        while(running) {
            {
                epicsGuardRelease<epicsMutex> U(G);
                wakeup.wait();
            }
            while(!work.empty()) {
                std::weak_ptr<...> wjob(work.front);
                work.pop_front();
                {
                    epicsGuardRelease<epicsMutex> U(G);
                    std::shared_ptr<...> job(wjob.lock());
                    try { if(job) job->whatever(); } catch ...
                }
            }
        }
    }
```

The key points being that the lock is held in `run()` by default, and only released when specifically needed.  Also, strong references to the work to be executed only exist in the same scope as the `epicsGuardRelease`.  I put this outside of the innermost try/catch since I often want to use some aspect in a log message.

anjohnson:
Assuming the Appveyor builds don't fail, I think this is now done. I added a separate test program for the notifyConveyor, and restructured the `run()` method as you suggested; I gave up trying to allow a client to be able to delete the conveyor in its `notifyClient()` method, if a client tries that now the notifier thread calls `cantProceed()`.

This PR should also fix #163 and #83.

mdavidsaver:
The 3.15 DLL builds fail with:

```
dbdToPv.obj : error LNK2001: unresolved external symbol dbr_value_offset
dbdToPv.obj : error LNK2001: unresolved external symbol dbf_text_dim
pvAccessCA.dll : fatal error LNK1120: 2 unresolved externals
```

wrt.

```
... fatal error LNK1318: Unexpected PDB error; OK (0) ...
```

I've yet to find a good explanation for what is happening.  [LNK1318](https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-error-lnk1318?view=msvc-160) seems to mean something went wrong with a (the?) local debug symbol server.  Given the intermittent occurrences, one "resolution" stands out to me:

> * Serialize linking to mitigate parallel link issues if needed. This error can be caused if mspdbsrv.exe is launched by one instance of link, and is shut down before another instance of link is done using it. The downside to this fix is that your project builds may take considerably longer to complete.

Maybe this could be avoided by preemptively launching a single instance?  (MSVC debug symbol handling is so complicated...)

mdavidsaver:
The appveyor build passed once.  I've restarted it, and will likely so a couple of more times, in an attempt to provoke a hang.  If not, then I'll merge.

anjohnson:
This build failure occurred while linking `testGetPerformance.exe`:
```
[00:08:01] make[2]: Entering directory 'C:/projects/pvaccesscpp/testApp/O.windows-x64-static'
[00:08:01] link -nologo  -debug -incremental:no -fixed:no -MACHINE:X64             -out:testGetPerformance.exe   testGetPerformance.obj      ../../lib/windows-x64-static/pvAccess.lib  C:/Users/appveyor/.cache/pvdata-master/lib/windows-x64-static/pvData.lib  C:/Users/appveyor/.cache/base-7.0/lib/windows-x64-static/Com.lib netapi32.lib netapi32.lib ws2_32.lib ws2_32.lib advapi32.lib user32.lib kernel32.lib winmm.lib dbghelp.lib     
[00:08:01] pvData.lib(createRequest.obj) : warning LNK4204: 'C:\projects\pvaccesscpp\testApp\O.windows-x64-static\vc140.pdb' is missing debugging information for referencing module; linking object as if no debug info
...
[00:08:01] Com.lib(osdSockAddrReuse.obj) : warning LNK4204: 'C:\projects\pvaccesscpp\testApp\O.windows-x64-static\vc140.pdb' is missing debugging information for referencing module; linking object as if no debug info
[00:08:01] Com.lib(epicsString.obj) : fatal error LNK1318: Unexpected PDB error; OK (0) 'C:\projects\pvaccesscpp\testApp\O.windows-x64-static\vc140.pdb'
[00:08:01] make[2]: *** [C:/Users/appveyor/.cache/base-7.0/configure/RULES_BUILD:212: testGetPerformance.exe] Error 1318
[00:08:01] make[2]: Leaving directory 'C:/projects/pvaccesscpp/testApp/O.windows-x64-static'
```

In an earlier failed build the same random error has happened while linking `testPVUnion.exe` which is in pvData, so I'm not convinced this has anything to do with caProvider as such. We could try using the `-Z7` compiler flag instead of `-Zi` which embeds the debugging symbols into the object files instead of using the separate shared database, but that flag belongs in base/configure/os/CONFIG.win32-x86.win32-x86 and not in the pvAccess module at all. This might solve both the LNK4204 and LNK1318 messages.

Thanks for re-triggering the build just now, I still don't have the ability to manage/configure the projects in the epics-base account.

mdavidsaver:
> I'm not convinced this has anything to do with caProvider as such.

Well I am convinced that it has nothing whatsoever to do with caProvider.  I see this error occasionally with other projects.  It just seems to happen more often with pvAccessCPP.  Go figure.  Another CI nuisance which will eventually be sorted out.

anjohnson:
We appear to have switched between `-Zi` and `-Z7` and back at least once in the past, so maybe it wouldn't help.

mdavidsaver:
If you are looking for something to try, [`/FS`](https://docs.microsoft.com/en-us/cpp/build/reference/fs-force-synchronous-pdb-writes?view=msvc-160) might be interesting.  In truth though, this particular issue doesn't seem so significant to me.  It's easy to identify as a build "failure".  I'll be more interested to get some details about whatever makes tests ~randomly crash on exit.

mdavidsaver:
Four runs with no hang.  That's good enough for now.

anjohnson:
Thanks.

We already use `-FS` on the compiler command-line for debug builds, which is where the failures have been happening:

`cl -EHsc -GR -DNOMINMAX -nologo -FC -D__STDC__=0 -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE -RTCsu -Zi -FS   -W3 -w44355 -w44344 -w44251    -D_SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING -MTd  -TP` ...

The error is coming from the linker and I don't see any relevant flags for that.

mdavidsaver:
> fatal error LNK1318: Unexpected PDB error; OK (0)

I've been paying more attention to these failures in the past week.  I've only seen this error with VS 2017 builds.  Not 2015 nor 2019.  Maybe this really is a toolchain bug?

anjohnson:
Possibly, [this might confirm that](http://masm32.com/board/index.php?topic=6571.0). They all seem to be 2017 static debug builds; maybe we should just exclude that particular combination?

Issue #150, opened 2019-07-15
RTEMS: race between recvfrom()/accept() and close() results in stuck threads


mdavidsaver:
Running testChannelAccess on RTEMS (for the first time?) leaves worker threads hung in recvfrom() and accept().  On investigation I find that this results if a socket is close()d while concurrent accept() or recvfrom() is in progress.

mdavidsaver:
The current shutdown sequence:

```c++
shutdown(socket, SHUT_RDWR);
close(socket);
thread.exitWait()
```

The problem with RTEMS can be avoided with:

```c++
shutdown(socket, SHUT_RDWR);
thread.exitWait()
close(socket);
```

However, this presents problems on targets other than Linux and RTEMS where shutdown() doesn't interrupt either accept() or recvfrom().

cf. https://github.com/epics-base/epics-base/pull/30

PR #169, opened 2020-12-02
RTEMS socket close() hack

mdavidsaver:
RTEMS allows blocking sockets to be interrupted by `shutdown()` (aka. `esscimqi_socketBothShutdownRequired`).  However, a concurrent `close()` is a race which can leave a `send()`/`recv()` call permanently stuck!  The _right_ way to handle this (aside from fixing the stack) would be to sequence `shutdown()` -> `exitWait()` -> `epicsSocketDestroy()`.  This is hard to handle since this must be done for _both_ sender and receiver thread, which presents difficulties owing to how the `Transport` hierarchy is structured since `Transport::close()` can happen on either worker, or a user thread.

Rather than try to straighten this mess out properly, we add this "wait" in-between `shutdown()` and `epicsSocketDestroy()` to hopefully wait for the workers (or other worker) to return from `send()`/`recv()`.

cf. #150

### DIFF
--- a/.github/workflows/ci-scripts-build.yml
+++ b/.github/workflows/ci-scripts-build.yml
@@ -75,14 +75,12 @@ jobs:
             configuration: default
             base: "7.0"
             rtems: "4.10"
-            test: NO
 
           - os: ubuntu-20.04
             cmp: gcc
             configuration: default
             base: "7.0"
             rtems: "4.9"
-            test: NO
 
           - os: ubuntu-16.04
             cmp: gcc-4.8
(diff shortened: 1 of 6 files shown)


mdavidsaver:
So far this problem has not manifest for (or hasn't been reported by) users.  However, it does appear if the remote/ unit tests are run for RTEMS, and results in hung tests.

I don't _think_ this will have ill effects for actual applications/IOCs.

I'm also disabling `testChannelAccess` for RTEMS as it has two failures which I'm not going to investigate at present.

```
   testChannelAccess.tap ..... 
  not ok 22 - void ChannelAccessIFTest::test_channel(): channel state change count should increase on the destroyed channel
  not ok 25 - void ChannelAccessIFTest::test_channel(): a destroy event was caught for the testing channel that was destroyed twice 
  Failed 4/152 subtests 
  	(less 12 skipped subtests: 136 okay)
```

mdavidsaver:
Merging now so that CI passes when I update the modules/pvAccess in Base.

PR #172, opened 2021-01-26
Enable building of pvtools to all except vxWorks, RTEMS and iOS.

hhslepicka:
EPICS Base tools build for host and also cross targets that are not vxWorks, RTEMS and iOS (https://github.com/epics-base/epics-base/blob/7.0/modules/ca/src/tools/Makefile#L16-L19). pvAccessCPP pvtools build only for host.

This PR makes so the pvAccessCPP pvtools follow the same standard.

This was tested at SLAC using a couple of cross targets that included RTEMS and linuxRT.
The tools were available for linuxRT and not RTEMS as expected.

### DIFF
--- a/pvtoolsSrc/Makefile
+++ b/pvtoolsSrc/Makefile
@@ -5,32 +5,36 @@ include $(TOP)/configure/CONFIG
 USR_CPPFLAGS += -I$(TOP)/src/utils
 USR_CPPFLAGS += -I$(TOP)/src/remote
 
-PROD_HOST += pvget
+PROD_DEFAULT += pvget
 pvget_SRCS += pvget.cpp
 pvget_SRCS += pvutils.cpp
 
-PROD_HOST += pvmonitor
+PROD_DEFAULT += pvmonitor
 pvmonitor_SRCS += pvmonitor.cpp
 pvmonitor_SRCS += pvutils.cpp
 
-PROD_HOST += pvput
+PROD_DEFAULT += pvput
 pvput_SRCS += pvput.cpp
 pvput_SRCS += pvutils.cpp
 
-PROD_HOST += pvcall
+PROD_DEFAULT += pvcall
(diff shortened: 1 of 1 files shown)


anjohnson:
This change looks right to me with the current build system. Base should probably have a way to make this easier so Makefiles like this wouldn't have to list all the OSs that _shouldn't_ build command-line programs. This is how ca/tools does it too, so I would accept this change.

hhslepicka:
@anjohnson I heavily based this change on the ca/tools Makefile, aka copied it.

Thank you for the time and review.

mdavidsaver:
@anjohnson Is the HOST vs IOC distinction still meaningful at this point?

anjohnson:
> Is the HOST vs IOC distinction still meaningful at this point?

Yes, but the terminology is a bit confusing. Only arch's with `Host` in `VALID_BUILDS` should install Perl scripts and build the `*_HOST` targets. Cross-builds don't because those targets aren't needed, e.g.:

```
tux% ls bin/linux-aarch64/
acctst*       caget*	  caput*	       casw*	 softIocPVA*
caConnTest*   cainfo*	  caRepeater*	       catime*
caEventRate*  camonitor*  caRepeater.service*  softIoc*

tux% ls bin/linux-x86_64/
acctst*		      dbdToMenuH.pl*	    mkmf.pl*
antelope*	      dbdToRecordtypeH.pl*  msi*
ascheck*	      dbExpand.pl*	    munch.pl*
assembleSnippets.pl*  depclean.pl*	    p2p*
... etc.
```

We don't have any host-only arch's any more, but AFAIK they should still be supported by the build system.

The thing we don't have is a set of variables such as `PROD_CMD` say for targets that should be built for cross-archs that can run regular command-line programs but which aren't build tools. We would also need a way to mark those architectures, and the `VALID_BUILDS` variable should be able to do that.

PR #174, opened 2021-03-09
Handle inputs with "=" to pvput

### LINKED ISSUE #168: pvput and CALC expressions
Trying to use pvput to set CALC field of a calcout record fails with a syntax that should be correct:
$ pvput timo:driveSeq.CALC "B<=A"
Old : <undefined>              B<A?VAL+1:0 INVALID DRIVER UDF 
**B< : Warning: no such field
Error: invalid bit-set length**
New : <undefined>              B<A?VAL+1:0 INVALID DRIVER UDF  

The same syntax works with caput though:

$ caput timo:driveSeq.CALC "B<=A"
Old : timo:driveSeq.CALC             B<A?VAL+1:0
New : timo:driveSeq.CALC             B<=A

A bit of a fringe case but I thought to report this nevertheless. I try to use pvtools for all my command line work.

jesusvasquez333:
This PR resolves #168 

The argument handling code of the `pvput` tool was reworked in order to handle bare value with `=` characters, which could be used for example to write to `DESC` or `CALC` fields. 

When trying to tread an input argument as a `field=value` pair, the code verifies if the `field` exists in the remote PV structure. If it doesn't exist, then it checks if the `.value` field is of type `string` and if it is, then the whole input value is treated as a bare value; otherwise the input is ignored (due to the `filed` being invalid).

### DIFF
--- a/pvtoolsSrc/pvput.cpp
+++ b/pvtoolsSrc/pvput.cpp
@@ -90,12 +90,7 @@ struct Putter : public pvac::ClientChannel::PutCallback
 
     Putter() :done(false) {}
 
-    typedef pvd::shared_vector<std::string> bare_t;
-    bare_t bare;
-
-    typedef std::pair<std::string, std::string> KV_t;
-    typedef std::vector<KV_t> pairs_t;
-    pairs_t pairs;
+    std::vector<std::string> values;
 
     pvd::shared_vector<std::string> jarr;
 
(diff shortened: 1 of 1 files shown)


pvtoolsSrc/pvput.cpp
```diff
@@ -333,34 +376,11 @@ int main (int argc, char *argv[])
...
-            fprintf(stderr, "\nCan't mix bare values and field=value pairs\n");
-            return 1;
-
-        } else if(thework.bare.size()==1 && thework.bare[0][0]=='[') {
-            // treat plain "[...]" as "value=[...]"
-            thework.pairs.push_back(std::make_pair("value", thework.bare[0]));
-            thework.bare.clear();
-        }
+        // copy values from command line
+        for (int n = 0; optind < argc; n++, optind++)
```
mdavidsaver, line 382: minor fyi.  The unused `n` could be omitted.  `for(; cond; var++) {}` is perfectly valid.

mdavidsaver:
Looks reasonable.

pvtoolsSrc/pvput.cpp
jesusvasquez333: Oh yes, that's right. I didn't notice `n` was unused in this case. Thanks!

PR #175, opened 2021-07-29
Fix for CA put using PV structures that do not have subfield "value"

sveseli:
The problem was described here: https://github.com/epics-base/pvaPy/issues/61

### DIFF
--- a/src/ca/dbdToPv.cpp
+++ b/src/ca/dbdToPv.cpp
@@ -830,6 +830,9 @@ Status DbdToPv::putToDBD(
     dbr_double_t dvalue(0);
     if(isArray) {
        PVScalarArrayPtr pvValue = pvStructure->getSubField<PVScalarArray>("value");
+       if(!pvValue) {
+           return Status(Status::STATUSTYPE_ERROR, string("DbdToPv::putToDBD logic error"));
+       }
        switch(caValueType) {
            case DBR_STRING:
            {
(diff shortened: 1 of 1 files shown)


mdavidsaver:
@anjohnson Can you have a look?

At a glance, there are a number of other places in dbdToPv.cpp where calls to `getSubField()` are missing `NULL` checks.

https://github.com/epics-base/pvAccessCPP/blob/a5cae7ad9242d099ee0602cb3061987e1ff95ef4/src/ca/dbdToPv.cpp#L462-L463

https://github.com/epics-base/pvAccessCPP/blob/a5cae7ad9242d099ee0602cb3061987e1ff95ef4/src/ca/dbdToPv.cpp#L585-L588

https://github.com/epics-base/pvAccessCPP/blob/a5cae7ad9242d099ee0602cb3061987e1ff95ef4/src/ca/dbdToPv.cpp#L623-L624

mdavidsaver:
Also note that with this change, testCaProvider is failing with:

```
  testCaProvider.tap .. 
  not ok 132 - waitPut(DBRmbbout) succeeded
  not ok 138 - waitPut(DBRbinaryout) succeeded
```

anjohnson:
Thanks; I've already been chatting with @sveseli who is planning to fix the failing tests. I will take a closer look for other similar bugs once he's finished doing that and this PR is passing.

sveseli:
The PR does not seem to see the most recent commit with checks for null pointer. Per @anjohnson, I will try closing/reopening it.

anjohnson (approved):
I'm pretty sure @sveseli has fixed all the places where `getSubField<>()` is called in this PR, I found each one in the original and matched them with the changes here. The phrase "logic error" doesn't seem quite right for all the error messages though, I might replace it with something a bit more descriptive when I merge this.

PR #176, opened 2021-10-28
CA Provider speed-up when connecting tens of thousands of CA channels

anjohnson:
This uses Jeff Hill's tsDLList implementation in the CA Provider to keep track of all the currently active CA Channel objects. I couldn't find a C++ standard container which didn't slow down the creation or deletion of channels or have unbounded memory issues (or require a garbage collection background thread). Maybe a `std::map` might have worked and been acceptable, but it would still have been slower than this.

Our user's application went from over 70 seconds to connect 60,000 PVs with EPICS 7.0.6 to under 3 with this version.

### DIFF
--- a/documentation/release_notes.dox
+++ b/documentation/release_notes.dox
@@ -1,5 +1,13 @@
 /** @page pvarelease_notes Release Notes
 
+Release 7.1.6 (UNRELEASED)
+=========================
+
+- Changes to caProvider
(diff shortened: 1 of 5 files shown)


sveseli (approved):
It would be great if tests for ca provider included use case where multiple channels are being connected/disconnected.

Issue #177, opened 2021-11-19
polluting ostream flags


dirk-zimoch:
Some functions/methods/operators manipulate the flags of their `std::ostream` argument  for examples using `std::hex` and/or `std::setw` without saving the initial `flags()` and restoing it when done and thus polluting the state of the ostream.

Examples are:
- HexDump <<
- ResponseHandler::handleResponse
- ClientResponseHandler::handleResponse
- ServerResponseHandler::handleResponse

Also affected but in pvData:
- shared_ptr_base::show_stack

Be aware that `<<std::hex` not only affect the next value but all future values.

The solution is to save and restore the flags:
```
std::ios::fmtflags initialflags = output.flags();
output << std::hex() << ...
output.flags(initialflags);
```

dirk-zimoch:
Correction: `setw()` resets after the next argument. But not `hex` or `setfill()`.

anjohnson:
I'm going to assume the merge of #178 fixed this issue, please reopen if there were parts not fixed by that.

PR #178, opened 2021-11-19
reset ostream flags when done

dirk-zimoch:
Fix for issue #177.
Always reset `flags()` when leaving a function that modifies them.
Also, setting `hex` or `setfill()` once is enough.


mdavidsaver:
Strictly speaking, the [fill character](https://en.cppreference.com/w/cpp/io/basic_ios/fill) should also be restored.  Still, this is a step in the right direction.

PR #179, opened 2022-02-19
avoid dangling pointer when indexvalue does out of scope

xiaoqiangwang:
The problem shows up on macOS but not on Linux and Windows as I checked.

With caProvider, the value will not change.
```
% pvput -p ca catest.SCAN 1
Old : 2022-02-19 15:25:35.846  MINOR DEVICE HIGH (0) Passive
New : 2022-02-19 15:25:35.846  MINOR DEVICE HIGH (0) Passive
```
With pvaProvider, the value is changed.
```
% pvput -p pva catest.SCAN 1
Old : 2022-02-19 15:25:35.846  MINOR DEVICE HIGH (0) Passive
New : 2022-02-19 15:25:35.846  MINOR DEVICE HIGH (1) Event
```

### DIFF
--- a/src/ca/dbdToPv.cpp
+++ b/src/ca/dbdToPv.cpp
@@ -840,6 +840,7 @@ Status DbdToPv::putToDBD(
     char *ca_stringBuffer(0);
     dbr_char_t   bvalue(0);
     dbr_short_t  svalue(0);
+    dbr_enum_t   evalue(0);
     dbr_long_t   lvalue(0);
     dbr_float_t  fvalue(0);
     dbr_double_t dvalue(0);
@@ -944,8 +945,7 @@ Status DbdToPv::putToDBD(
            case DBR_ENUM:
            {
                GET_SUBFIELD_WITH_ERROR_CHECK(PVInt, pvValue, pvStructure, "value.index", "DbdToPv::putToDBD logic error");
-               dbr_enum_t indexvalue = pvValue->get();
-               pValue = &indexvalue;
+               pValue = put_DBRScalar<dbr_enum_t,PVUShort>(&evalue,pvValue);
                break;
            }
            case DBR_STRING: 


anjohnson:
I can confirm this fixes pvput to a SCAN field using the caProvider on MacOS, thanks!

Unfortunately adding a test case to the testCa/testCaProvider.cpp code to detect this doesn't seem trivial to do, the test output showed that something was wrong without this fix when I made this change:
```patch
diff --git a/testCa/testCaProvider.cpp b/testCa/testCaProvider.cpp
index 30ce61ce..8703239f 100644
--- a/testCa/testCaProvider.cpp
+++ b/testCa/testCaProvider.cpp
@@ -833,6 +833,7 @@ void testClientTypes(void)
     checkClient("DBRdoubleArray", "1 2 3");
     checkClient("DBRstringArray", "aa bb cc");
     checkClient("DBRmbbout", "2");
+    checkClient("DBRmbbout.SCAN", "4");
     checkClient("DBRbinaryout", "1");
 }
 
@@ -855,7 +856,7 @@ MAIN(testCaProvider)
 #if HAS_DBUNITTEST
         // Set TESTCAP_USE_SYSTEM in environment to use other impl.
         if (!getenv("TESTCAP_USE_SYSTEM")) {
-            testPlan(143);
+            testPlan(149);
             TEST_CLIENT_TYPES(TestIocUnit)
         } else
 #endif
```
but it didn't cause any of the actual tests on that channel to fail:
```
# TestClient::put DBRmbbout.SCAN := 4
 filename="(null)" line number=0PV: DBRmbbout scanAdd detected illegal SCAN value
```

I am about to merge this PR (without the above test modifications).

I don't have time to rewrite testCaProvider.cpp at the moment, I think it probably needs that level of work to fix properly.

PR #182, opened 2022-05-15
Fix crash when toEcho is empty

FreddieAkeroyd:
I had some repeatable crashes locally on windows x64 with a debug build until I applied this fix, but having just reverted the fix to get a stack trace to post now I don't seem to be able to reproduce it. Just posting fix here in case it might be something worth applying anyway.

### DIFF
--- a/src/server/pv/responseHandlers.h
+++ b/src/server/pv/responseHandlers.h
@@ -96,7 +96,7 @@ public:
     EchoTransportSender(osiSockAddr* echoFrom, size_t payloadSize, epics::pvData::ByteBuffer& payloadBuffer) {
         memcpy(&_echoFrom, echoFrom, sizeof(osiSockAddr));
         toEcho.resize(payloadSize);
-        payloadBuffer.getArray(&toEcho[0], payloadSize);
+        payloadBuffer.getArray(toEcho.data(), payloadSize);
     }
 
     virtual ~EchoTransportSender() {}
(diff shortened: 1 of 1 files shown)


mdavidsaver:
I have a faint memory that some PVA peers send zero length echo messages.  Combined with your mention of "debug", I suspect the case where `payloadSize==0` is triggering an `assert()` failure because `data[0]` is strictly speaking out of range.  The `&data[0]` expression should be safe as it should not actually de-reference the invalid address.  There are several "should"s here because I (and others) am depending on the observation that all known implementations of `std::vector` use a single allocation.  However, I've hit problems in this area [before](https://github.com/epics-base/pvDataCPP/blob/f3911d583139ac5969df6a992c424788e41d9634/src/misc/pv/sharedVector.h#L547-L548) with MSVC debug builds.

Unfortunately, [`std::vector<T>::data()`](https://en.cppreference.com/w/cpp/container/vector/data) wasn't added until c++11 (otherwise I drop the `&data[0]` hack).

If my guess is correct, then one way to fix this would be to make the call to `getArray()` conditional on there actually being some bytes to copy out.  eg.

```cpp
if(payloadSize) {
    payloadBuffer.getArray(&toEcho[0], payloadSize);
}
```

FreddieAkeroyd:
yes, it was a debug assert on accessing a size 0 vector. Your suggested alternative should work equally well, I had forgot data() was a more recent addition .

anjohnson:
Core Group: Approved.

PR #185, opened 2022-10-27
Set PVAS_SERVER_PORT

mdavidsaver:
Sets `$PVAS_SERVER_PORT` for use with recceiver.

Note that this would also set `$PVAS_SERVER_PORT` each time a server is restarted, potentially with a different port.  Currently, there is no good way to notify reccaster of this change.


anjohnson:
Core Group: Approved.

PR #191, opened 2023-10-25
clientContextImpl: Cap the number and age of beacons

### LINKED ISSUE #184: Resource leak in Beacon tracking
The `m_beaconHandlers` map can grow without bound as new PVA servers come online.

https://github.com/epics-base/pvAccessCPP/blob/b2f0aecfa362a0616758d7aefe3006c21abe754d/src/remoteClient/clientContextImpl.cpp#L4525-L4530

This map needs to be bounded in size and/or time.  (like [PVXS does...](https://github.com/mdavidsaver/pvxs/blob/f22ab94458abb9ed58af0582e88cf1b2b40ed52a/src/client.cpp#L1084-L1121))

JJL772:
Each beacon has an associated mutex. If we don't cap the beacon count, IOCs running on resource limited platforms like RTEMS may eventually run out of resources and crash.

Closes #184

The configuration options are probably unnecessary, let me know if I should remove them.


src/remoteClient/clientContextImpl.cpp
```diff
@@ -4351,6 +4355,38 @@ class InternalClientContextImpl :
...
         if (it == m_beaconHandlers.end())
         {
+            TimeStamp now;
+            now.getCurrent();
+
+            double oldestAge = -1;
+            AddressBeaconHandlerMap::iterator oldestBeacon = m_beaconHandlers.end();
+
+            /* Before creating a new beacon, cleanup any old ones */
+            for (AddressBeaconHandlerMap::iterator it = m_beaconHandlers.begin(); it != m_beaconHandlers.end();)
```
mdavidsaver, line 4365: This is a linear iteration of the list of tracked servers each time a beacon is received.  If the number of servers grows large, or one gets stuck in a restart loop, this could potentially take up a lot of CPU time.

`Timer InternalClientContextImpl::m_timer` is available in `getBeaconHandler()`.  cf. `class ChannelSearchManager` for an example of using `Timer`.

src/remoteClient/clientContextImpl.cpp
```diff
@@ -3957,7 +3957,7 @@ class InternalClientContextImpl :
     static size_t num_instances;
 
     InternalClientContextImpl(const Configuration::shared_pointer& conf) :
-        m_addressList(""), m_autoAddressList(true), m_connectionTimeout(30.0f), m_beaconPeriod(15.0f),
+        m_addressList(""), m_autoAddressList(true), m_connectionTimeout(30.0f), m_beaconPeriod(15.0f), m_maxBeacons(10), m_maxBeaconLifetime(15),
```
mdavidsaver, line 3960: A limit of 10 servers, and a 15 second lifetime seem quite small.  Have you checked one a network with more than 10 PVA servers to see how often `ChannelSearchManager::newServerDetected()` is called?

src/remoteClient/clientContextImpl.cpp
```diff
@@ -4115,6 +4117,8 @@ class InternalClientContextImpl :
         m_beaconPeriod = m_configuration->getPropertyAsFloat("EPICS_PVA_BEACON_PERIOD", m_beaconPeriod);
         m_broadcastPort = m_configuration->getPropertyAsInteger("EPICS_PVA_BROADCAST_PORT", m_broadcastPort);
         m_receiveBufferSize = m_configuration->getPropertyAsInteger("EPICS_PVA_MAX_ARRAY_BYTES", m_receiveBufferSize);
+        m_maxBeacons = m_configuration->getPropertyAsInteger("EPICS_PVA_BEACON_MAX", m_maxBeacons);
+        m_maxBeaconLifetime = m_configuration->getPropertyAsFloat("EPICS_PVA_BEACON_LIFETIME", m_maxBeaconLifetime);
```
mdavidsaver, line 4121: I can see making the number of servers configurable.  However, I don't think the lifetime (aka. 2x the max beacon period of 180 seconds) needs to be separately configurable.

mdavidsaver (changes requested):
I am happy to see a PR addressing this issue.  I think some further work is required though.

mdavidsaver:
Since it is not super obvious.  The Beacon TX timing of pvAccessCPP differs from what experience with CA servers might lead you to expect.

https://github.com/epics-base/pvAccessCPP/blob/581d100a029e6f68f3c1830b90509b3ee5ebd7a0/src/server/beaconEmitter.cpp#L31-L33

A server will send out the first 10 beacons (not configurable) with a 15 second interval (by default), then switch to a 180 second period (not configurable).  While `$EPICS_PVAS_BEACON_PERIOD` can override the first "fast" period, I don't think this is in practice useful.

So I think the beacon tracking lifetime must be >= 360 seconds.

(fyi. with PVXS I try to follow the same model and timings, with a non-configurable limit of 20k servers.  Of course, there I only allocate ~64 bytes per server)

mdavidsaver:
The windows CI failures are due to https://github.com/epics-base/ci-scripts/issues/84.  When you update, please rebase to pick up ed7eae59bea24645e18562e5238e281fd12e5a99.

JJL772:
@mdavidsaver Thanks for the feedback! I finally got around to applying the requested changes.
I ended up using pvxs as a reference and copied the max beacon lifetime (360s) and beacon limit (20000).

JJL772:
@mdavidsaver Just wanted to follow up, are there any other changes required for this?

src/remoteClient/clientContextImpl.cpp
```diff
@@ -4351,8 +4385,18 @@ class InternalClientContextImpl :
...
+            {
+                char ipa[64];
+                sockAddrToDottedIP(&responseFrom->sa, ipa, sizeof(ipa));
+                LOG(logLevelDebug, "Tracked beacon limit reached (%d), ignoring %s\n", maxTrackedBeacons, ipa);
+                return BeaconHandler::shared_pointer();
+            }
+
             // stores weak_ptr
             handler.reset(new BeaconHandler(internal_from_this(), responseFrom));
+            m_timer->scheduleAfterDelay(BeaconCleanupCallback::shared_pointer(new BeaconCleanupCallback(*this, *responseFrom)), maxBeaconLifetime);
```
mdavidsaver, line 4399: `class Timer` can be a little strange to use.  As written, this timer will never be `cancel()`ed.  If fact, since the callback pointer isn't be stored, it can't be canceled.  So each time a new server appears, this timer will drop it 360 seconds later.  Then, since that server will probably still be active, it will be "new" again 180 after than.

I think you are on the right track though.  What you would need to do is to store a `BeaconCleanupCallback` in (or as part of) `BeaconHandler`.  Then the logic would be to start the timer on creation (first beacon).  Then cancel and restart it on each subsequent beacon.  This way the timer will only expire when the corresponding server stops sending beacons.

mdavidsaver (changes requested):

JJL772:
This PR now depends on some changes made to pvData: https://github.com/epics-base/pvDataCPP/pull/94 

I'm going to mark this as a draft for now because I'm not exactly happy with these changes yet.

src/remoteClient/clientContextImpl.cpp
```diff
@@ -4556,8 +4608,44 @@ class InternalClientContextImpl :
...
+{
+    m_impl.m_timer->cancel(m_callback);
+}
+
+void BeaconCleanupHandler::Callback::callback()
+{
+    epicsTimeStamp ts;
+    epicsTimeGetCurrent(&ts);
+    if (epicsTimeDiffInSeconds(&ts, &m_handler.m_lastTime) > maxBeaconLifetime)
+        m_handler.remove();
```
mdavidsaver: I don't see any need for precision here.  And calling `epicsTimeGetCurrent()` potentially 20,000 times could be avoided.

Replace `epicsTimeStamp m_lastTime` with a integer counter.  Set the timer to eg `maxBeaconLifetime/2`.  Increment the counter on expiration.  If it reaches 3, then `remove()`.  Have `touch()` zero the counter.

That said, this is an optimization.  What you have here will work.  So please let me know if you will not have to for this change.

mdavidsaver: I don't have a good feel for what this cap should be.  My most recent solid point of reference is NSLS2 circa 2015, when I snooped on beacon traffic to identify ~1000 CA server instances running on ~100 hosts.  Well short of 20k, but only a factor of 2 from 2048.

PVXS currently tracks up to 20k addresses.  I picked this number from thin air.  Although I opted for a larger number because PVXS uses fewer resources per server ( ~= 256 bytes per).

mdavidsaver (approved):
Looks ok.  Thanks for your persistence!

src/remoteClient/clientContextImpl.cpp
anjohnson: Assuming this cap is the maximum number of servers that could be generating beacons visible to a client, the APS accelerator currently has 1136 IOCs in production (mostly CA currently) spread across 5 VLANs, with 881 IOCs in the largest VLAN. I don't see our total number increasing beyond 2048 IOCs, but we do have 13 VMs that run multiple soft IOCs (the largest VM has 214) so the number of visible servers isn't limited to the subnet size. Presumably configuring PVA servers through a shared multicast address would include the servers from multiple subnets too.

I'm fine with the cap being 2048 for pvAccessCPP. It might be worth having the PVXS cap be configurable by an env-var at some point, how many IOCs is ITER expecting to run?

Issue #159, opened 2020-03-18
PVA Server's use of random 'sendTransport' UDP port makes it impractical for gateway


kasemir:
One key use case for a CA or PVA gateway is enabling read access to the control system through firewalls.
Inside the control system network, servers and clients use broadcasts to find each other.
From the outside, only very few UDP and TCP ports are open, limiting access to just a gateway.

When using the P4P gateway, which is based on the pvAccessCPP server, such a setup is nearly impossible because the gateway's port usage can't be controlled.

When starting a pvAccessCPP server with  'EPICS_PVAS_SERVER_PORT=5077', the server will listen to clients on UDP 5077 and then also serve data via TCP 5077, so you can open UDP and TCP 5077 on the firewall. But whenever the server receives a client's search request on UDP 5077, the server sends the "I have that channel, talk to me on TCP 5077" message from an arbitrary UDP port, not UDP 5077. With only UDP and TCP 5077 allowed through the firewall, the client never sees the reply. On option to make this work is configuring the firewall to allow all outgoing UDP traffic from the server, but that's counter the purpose of a firewall.

The reason for this  pvAccessCPP server behavior is in the setup of its `sendTransport`.
With a very simple patch, the server would use the configured UDP port for the `sendTransport`, so search requests received by the configured port result in a reply from that same port, firewall configuration is easy and based on preliminary tests it all "works":

```
diff --git a/src/remote/blockingUDPTransport.cpp b/src/remote/blockingUDPTransport.cpp
index 16b377a..d25b161 100644
--- a/src/remote/blockingUDPTransport.cpp
+++ b/src/remote/blockingUDPTransport.cpp
@@ -594,7 +594,7 @@ void initializeUDPTransports(bool serverFlag,
     osiSockAddr anyAddress;
     memset(&anyAddress, 0, sizeof(anyAddress));
     anyAddress.ia.sin_family = AF_INET;
-    anyAddress.ia.sin_port = htons(0);
+    anyAddress.ia.sin_port = htons(listenPort);
     anyAddress.ia.sin_addr.s_addr = htonl(INADDR_ANY);
 
     sendTransport = connector.connect(responseHandler, anyAddress, protoVer);
```

There may, however, be dragons. Issue #128 mentions problems with sharing UDP ports. Michael cautioned negative side effects related to the CMD_ORIGIN_TAG handling, resulting in message loops.

[...]

PR #193, opened 2024-01-25
UDP Sender Port

sveseli:
This PR allows control of UDP sender port on both client and server side via the EPICS_PVA_UDP_SENDER_PORT environment variable. If this is not set, the behavior on both client and server side is as before, with random port assigned for UDP sender messages.

Closes https://github.com/epics-base/pvAccessCPP/issues/159 .


sveseli:
The proposed changes will be merged with #192.

PR #195, opened 2024-04-16
add default message for pvget

rerpha:
Adds a default message for pvget. I don't know if line 363 (the default case) was a bug as i could never get it to run.  @ralphlange I found this after I showed you my (accidentally?) working PVAccess. 

Maybe there's a good reason that pvget doesn't have this line but the other tools do, please let me know!

### DIFF
--- a/pvtoolsSrc/pvget.cpp
+++ b/pvtoolsSrc/pvget.cpp
@@ -276,6 +276,12 @@ int MAIN (int argc, char *argv[])
 
         // ================ Parse Arguments
 
+        if(argc <= optind){
+            /* No arguments specified */
+            fprintf(stderr, "No arguments specified. Please use " EXECNAME" -h for help");
+            return 1;
+        }
+
         while ((opt = getopt(argc, argv, ":hvVRM:r:w:tmp:qdcF:f:ni")) != -1) {
             switch (opt) {
             case 'h':               /* Print usage */
@@ -361,6 +367,9 @@ int MAIN (int argc, char *argv[])
                         optopt);
                 return 1;
             default :
+                fprintf(stderr,
+                        "Option '-%c' is not supported - it is a valid option but is not implemented. \n",
+                        optopt);
                 usage();
                 return 1;
             }


mdavidsaver:
> I don't know if line 363 (the default case) was a bug as i could never get it to run.

That default case should be impossible so long as the `switch` is consistent with the string passed to `getopt()`.  imo. any inconsistency should be an error.  I usually combine default and `'?'` though.

> Maybe there's a good reason that pvget doesn't have this line but the other tools do, please let me know!

I guess you are referring to `caget`?

It think it is a bit of a judgement call whether PVs is an error with `pvget`.  With `pvmonitor` I think this should be.

Is it worth changing behavior at this point?

rerpha:
> > Maybe there's a good reason that pvget doesn't have this line but the other tools do, please let me know!
> 
> I guess you are referring to `caget`?
> 
> It think it is a bit of a judgement call whether PVs is an error with `pvget`. With `pvmonitor` I think this should be.
> 
> Is it worth changing behavior at this point?

by other tools yes caget, but also pvput, pvinfo, pvmonitor all seem to give the "No pv name(s) specified. ('pvXXXX -h' for help.)" message - just thinking for consistency among the pv* utils it might be nice.

pvtoolsSrc/pvget.cpp
```diff
@@ -360,9 +366,6 @@ int MAIN (int argc, char *argv[])
                         "Option '-%c' requires an argument. ('" EXECNAME " -h' for help.)\n",
                         optopt);
                 return 1;
-            default :
-                usage();
-                return 1;
```
anjohnson, line 365: Should this code retain the default case handling, to catch any future mis-matches between the options string passed into `getopt()` and the set of cases that the code actually handles? It might want to output a different error message though.

rerpha: I think the question mark will do that already?

anjohnson: I'm not sure whether you're answering my question or not, I'll re-phrase it. What happens if someone adds several new letters to the `getopt()` string but forgets to add a new `case` statement for one of those letters; or removes one or more cases but forgets to remove the option letter(s) from the string? The code has then told `getopt()` that it supports an option, but nothing actually handles that if a user tries it.

rerpha: ah sorry, yes i understand. I think printing something about the given option not being supported (despite what getopt thinks) and then showing the usage would make sense.

rerpha: have added back in https://github.com/epics-base/pvAccessCPP/pull/195/commits/b593e3da157751d642747e1d8327ef38bb373830

pvtoolsSrc/pvget.cpp
```diff
@@ -361,8 +367,11 @@ int MAIN (int argc, char *argv[])
                         optopt);
                 return 1;
             default :
-                usage();
-                return 1;
+                fprintf(stderr,
+                            "Option '-%c' is not supported - it is a valid option but is not implemented. \n",
+                            optopt);
+                    usage();
+                    return 1;
```
anjohnson, line 374: Please fix the indentation, these should line up with the fprintf() above.

rerpha: whoops, sorry, i forgot about this completely - now done.

anjohnson (approved):
Thank-you, this looks fine now.

Just a note for any future Pull Requests you make that have more complex changes and more than one commit: Please don't use `git pull` after committing changes, since that adds a merge commit to your PR branch. Our preferred method of moving your changes up to the latest commit on the 7.0 branch is to do a `git pull --rebase`. In this case though we will squash your changes into a single commit when we merge it so you don't need to do anything more here.

PR #201, opened 2025-02-28
Added json5 option to -M option for PVA tools

zhangt58:
- Explicitly request JSON5 format via `-M json5`
- While using `-M json` for the regular JSON string


zhangt58:
See also https://github.com/epics-base/pvDataCPP/pull/98

PR #202, opened 2025-05-05
Fix printf format warning

JJL772:
just a tiny warning fix.

### DIFF
--- a/src/ioc/reftrackioc.cpp
+++ b/src/ioc/reftrackioc.cpp
@@ -25,7 +25,7 @@ void showRefs(const epics::RefSnapshot& snap, int lvl, bool delta)
         if(it->second.current==0 && it->second.delta==0 && lvl<=0) continue;
         if(delta && it->second.delta==0 && lvl<=0) continue;
         if(delta) {
-            printf(" %s : %zu (delta %zd)\n",
+            printf(" %s : %zu (delta %ld)\n",
                    it->first.c_str(), it->second.current, it->second.delta);
         } else {
             printf(" %s : %zu\n", it->first.c_str(), it->second.current);


ralphlange:
LGTM.

The `size_t` problem remains: older Microsoft compilers don't know `%z`, but `size_t` on Windows is an `unsigned long long`.
In the past, casting to `unsigned long long` and using `%llu` was the only portable way.

Of course, C++ code could (should?) use `cout` instead.

mdavidsaver:
> older Microsoft compilers don't know `%z`

Is this still relevant with the currently supported versions?

Although, I think the real question is about runtime support.

> In the past, casting to `unsigned long long` and using `%llu` was the only portable way.

Personally, I have a small inclination to leave `"%z"` and continue to ignore these warnings.

Otherwise, a cast is necessary.

> Of course, C++ code could (should?) use `cout` instead.

This printf is in an iocsh function, where output redirection via `epicsStdio.h` is desired.

ralphlange:
> > older Microsoft compilers don't know `%z`
> Is this still relevant with the currently supported versions?

Ehh... Stackoverflow says `%z` and `%zu` are supported starting from VS2015.
We still have VS2013 in our AppVeyor list, but it will be a non-issue in the not-so-distant future.

> > Of course, C++ code could (should?) use `cout` instead.
> This printf is in an iocsh function, where output redirection via `epicsStdio.h` is desired.

Ups, yes, true.

JJL772:
`it->second.delta` is `long`: https://github.com/epics-base/pvDataCPP/blob/master/src/misc/pv/reftrack.h#L77

mdavidsaver:
> `it->second.delta` is `long`

So it is.  I am duly corrected.

anjohnson:
I'm currently updating the GHA jobs to Ubuntu-22. Will do the other original V4 submodules too, without PRs.

mdavidsaver:
> ... without PRs.

With test builds though, right?  ;)

anjohnson:
> test builds …

- [pvData Test](https://github.com/anjohnson/pvDataCPP/actions/runs/14842362635) => [Pushed](https://github.com/epics-base/pvDataCPP/actions/runs/14842714414)
- [pvAccess Test](https://github.com/anjohnson/pvAccessCPP/actions/runs/14842164248) => [Pushed](https://github.com/epics-base/pvAccessCPP/actions/runs/14842724669)
- [pvDatabaseCPP Test](https://github.com/anjohnson/pvDatabaseCPP/actions/runs/14292242057) => [Pushed](https://github.com/epics-base/pvDatabaseCPP/actions/runs/14842993158)
- [pva2pva Testing](https://github.com/anjohnson/pva2pva/actions) … WIP.

Neither the normativeTypes nor pvaClient modules have GHA configurations, and I don't intend to add them (at the moment at least).

PR #212, opened 2025-11-04
Fix MSVC warning C4275

dirk-zimoch:
MSVC warnings: non - DLL-interface class 'class_1' used as base for DLL-interface class 'class_2'

Fixes:
 * Make base classs`Destroyable` and `TransportSendControl` DLL-interface classes
 * Make `RPCRequestException` a non-DLL-interface class (does not need to be and should not be one if fully defined in header file, i.e. has no methods in *.cpp which would end up in the DLL)

### DIFF
--- a/src/remote/pv/remote.h
+++ b/src/remote/pv/remote.h
@@ -119,7 +119,7 @@ void hackAroundRTEMSSocketInterrupt();
 /**
  * Interface defining transport send control.
  */
-class TransportSendControl : public epics::pvData::SerializableControl {
+class epicsShareClass TransportSendControl : public epics::pvData::SerializableControl {
 public:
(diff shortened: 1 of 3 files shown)


dirk-zimoch:
> It should be sufficient to clone epics-base with all submodules, update this one, then build. If softIocPVA links, then all is well.

That's what I did. I do the hunt for warnings in the context of an epics-base build. softIocPVA.exe builds and starts without error.

BTW: I see the Win2019 builds failing. Could you update your CI config?

mdavidsaver:
> BTW: I see the Win2019 builds failing. Could you update your CI config?

This is on my TODO list.  If you are motivated, the change should be equivalent to epics-base/epics-base@2f0fbbe1c6c424e377faf7e2a560618b1bfbbd9c .

PR #218, opened 2025-11-21
Fix build errors and warnings from #184 

JJL772:
Fixes a couple of oopses in #184. I mistakenly used C++11 constructs (nullptr, std::shared_ptr) which breaks older compilers.

There was also a warning due to linkage type mismatch in the `BeaconHandler`, so I've moved `BeaconCleanupHandler` to a non-anonymous `detail` namespace


JJL772:
Ah, I misunderstood your previous comment. Feel free to test away

PR #226, opened 2026-07-10
Add protection against some network issues

anjohnson:
Moved from the private area.

By commit:

1. Reject negative payload sizes in messages over UDP and TCP, with tests.
It looks like this module can't handle TCP or UDP messages over 2^31 Bytes, although the TCP code apparently understands joining segmented messages.

2. Ignore over-long channel names in search messages.
This commit skips over names longer than would be accepted by the channel create code. A better response might be to immediately discard the search message, research needed on how to implement that.

3. Uncomment Lock guards that were improperly disabled long ago.
See [the gist here (2nd document)](https://gist.github.com/anjohnson/9cd360c3d1dc010aaddec46439a73838#file-lock-guard-analysis-md) for Claude's assessment of the lock guard history.

4. Prevent deadlocks in `destroy()` methods.
See [the gist here (1st document)](https://gist.github.com/anjohnson/9cd360c3d1dc010aaddec46439a73838#file-deadlock-check-serverchannelputgetrequesterimpl-send-md) for Claude's explanation of fixes like one by @mdavidsaver in 2018.

5. Limit strings accepted as channel names to `MAX_CHANNEL_NAME_LENGTH`, ignoring anything larger. This requires a change implemented in my pvData add-protection branch, epics-base/pvDataCPP#107

### DIFF
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -206,7 +206,8 @@ void ClientChannel::removeConnectListener(ConnectCallback* cb)
 void ClientChannel::show(std::ostream& strm) const
 {
     if(impl) {
-        strm<<typeid(*impl->channel.get()).name()<<" : ";
+        pva::Channel& channel = *impl->channel;
+        strm<<typeid(channel).name()<<" : ";
         impl->channel->printInfo(strm);
     } else {
         strm<<"NULL Channel";
@@ -346,8 +347,9 @@ void ClientProvider::disconnect()
 ::std::ostream& operator<<(::std::ostream& strm, const ClientChannel& op)
 {
     if(op.impl) {
+        pva::Channel& channel = *op.impl->channel;
         strm << "ClientChannel("
-             << typeid(*op.impl->channel.get()).name()<<", "
+             << typeid(channel).name()<<", "
                 "\"" << op.impl->channel->getChannelName() <<"\", "
                 "\"" << op.impl->channel->getProvider()->getProviderName() <<"\", "
                 "connected="<<(op.impl->channel->isConnected()?"true":"false")
@@ -361,8 +363,9 @@ void ClientProvider::disconnect()
 ::std::ostream& operator<<(::std::ostream& strm, const ClientProvider& op)
 {
     if(op.impl) {
+        pva::ChannelProvider& provider = *op.impl->provider;
         strm << "ClientProvider("
-             << typeid(*op.impl->provider.get()).name()<<", "
+             << typeid(provider).name()<<", "
                 "\""<<op.impl->provider->getProviderName()<<"\")";
     } else {
         strm << "ClientProvider()";
--- a/src/pipelineService/pv/pipelineService.h
+++ b/src/pipelineService/pv/pipelineService.h
@@ -34,7 +34,7 @@ class epicsShareClass PipelineControl
 public:
     POINTER_DEFINITIONS(PipelineControl);
 
-    virtual ~PipelineControl() {};
+    virtual ~PipelineControl();
 
     /// Number of free elements in the local queue.
     /// A service can (should) full up the entire queue.
@@ -64,7 +64,7 @@ class epicsShareClass PipelineSession
 public:
     POINTER_DEFINITIONS(PipelineSession);
 
-    virtual ~PipelineSession() {};
+    virtual ~PipelineSession();
 
     /// Returns (minimum) local queue size.
     /// Actual local queue size = max( getMinQueueSize(), client queue size );
@@ -88,7 +88,7 @@ class epicsShareClass PipelineService
 public:
     POINTER_DEFINITIONS(PipelineService);
 
-    virtual ~PipelineService() {};
+    virtual ~PipelineService();
 
     virtual PipelineSession::shared_pointer createPipeline(
         epics::pvData::PVStructure::shared_pointer const & pvRequest
--- a/src/remote/blockingUDPTransport.cpp
+++ b/src/remote/blockingUDPTransport.cpp
@@ -369,6 +369,10 @@ bool BlockingUDPTransport::processBuffer(Transport::shared_pointer const & trans
         if (flags & 0x01)
             continue;
 
+        // reject negative payload size (sign-extends to a huge size_t)
+        if (static_cast<int32>(payloadSize) < 0)
+            return false;
+
         size_t nextRequestPosition = receiveBuffer->getPosition() + payloadSize;
 
         // payload size check
--- a/src/remote/codec.cpp
+++ b/src/remote/codec.cpp
@@ -242,6 +242,13 @@ void AbstractCodec::processReadNormal()  {
                         "not-a-first segmented message received in normal mode");
                 }
 
+                // reject negative payload size (sign-extends to a huge size_t)
+                if (_payloadSize < 0)
+                {
+                    invalidDataStreamHandler();
+                    throw invalid_data_stream_exception("negative payload size");
+                }
+
                 _storedPayloadSize = _payloadSize;
                 _storedPosition = _socketBuffer.getPosition();
                 _storedLimit = _socketBuffer.getLimit();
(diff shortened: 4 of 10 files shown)


anjohnson:
Oops, that last commit needs a change from my [pvDataCPP add-protection](https://github.com/epics-base/pvDataCPP/pull/107) branch.

mdavidsaver:
> 1. Reject negative payload sizes in messages over UDP and TCP, with tests.
>        It looks like this module can't handle TCP or UDP messages over 2^31 Bytes, although the TCP code apparently understands joining segmented messages.

No problem with UDP.

I have some hesitation about TCP.  Have you instead tried to switch `_payloadSize` to uint32?

The semantics around segmented messages has never been clear to me.  Currently PVXS can receive segmented messages, but will never send them.  It does handle message length as unsigned 32-bit.

@kasemir Does core.pva do anything with the segmentation flags?

kasemir:
> @kasemir Does core.pva do anything with the segmentation flags?

There is code to detect a segmented message and assemble the pieces, but I don't think I've ever been able to check that against a PVA server which generates such messages.

https://github.com/ControlSystemStudio/phoebus/blob/51a9bf6e4f882b21830f94531ba795a42f915c58/core/pva/src/main/java/org/epics/pva/common/TCPHandler.java#L426

anjohnson:
> Have you instead tried to switch _payloadSize to uint32?

Why? Doing that would require much more complicated changes to the code and tests.

The PVA protocol Wiki [explicitly says](https://github.com/epics-base/pvAccessCPP/wiki/Protocol-Encoding#sizes) about sizes:

3. If the number of elements is less than 2^31-1, then the size MUST be encoded as an unsigned 8-bit integer with value 254, followed by a positive signed 32-bit integer indicating the number of elements.
4. Values greater than or equal to 2^31-1 are currently not implemented.

It does go on to suggest how larger sizes would be encoded if they were supported though. Have you implemented them in PVXS without updating the spec? Does the new Java implementation also support that?

src/client/client.cpp
```diff
@@ -206,7 +206,8 @@ void ClientChannel::removeConnectListener(ConnectCallback* cb)
 void ClientChannel::show(std::ostream& strm) const
 {
     if(impl) {
-        strm<<typeid(*impl->channel.get()).name()<<" : ";
+        pva::Channel& channel = *impl->channel;
```
anjohnson, line 209: Needs to call `get()` to get the type-id of the contained object. Ditto below.

Move to a non-hardening PR.

src/pipelineService/pv/pipelineService.h
```diff
@@ -34,7 +34,7 @@ class epicsShareClass PipelineControl
 public:
     POINTER_DEFINITIONS(PipelineControl);
 
-    virtual ~PipelineControl() {};
+    virtual ~PipelineControl();
```
anjohnson, line 37: Not hardening.

src/remote/blockingUDPTransport.cpp
anjohnson, line 365: Remove TODO comment.

src/remote/blockingUDPTransport.cpp
```diff
@@ -369,6 +369,10 @@ bool BlockingUDPTransport::processBuffer(Transport::shared_pointer const & trans
         if (flags & 0x01)
             continue;
 
+        // reject negative payload size (sign-extends to a huge size_t)
+        if (static_cast<int32>(payloadSize) < 0)
+            return false;
+
```
anjohnson, line 375: Core: Add an int32 payload size var above which is assigned from the buffer, then exclude negatives and assign to the size_t version (with no renames or casts below).

src/remote/codec.cpp
anjohnson, line 190: Do the –ve checks for payload size with the magicCode check on 193 below, not later on.

src/remote/codec.cpp
```diff
@@ -1664,7 +1686,10 @@ void BlockingServerTCPTransportCodec::authNZInitialize(const std::string& securi
     info->authority = securityPluginName;
 
     if (!plugin->isValidFor(*info))
+    {
         verified(pvData::Status::error("invalid security plug-in name"));
+        return;
+    }
```
anjohnson, line 1692: Complex situation. There's a parallel thread running here, is `return` after the `verified()` call enough to signal that too?

Maybe try having the client code send a bad value to see what happens.

src/remote/security.cpp
```diff
@@ -319,11 +319,12 @@ void AuthNZHandler::handleResponse(osiSockAddr* responseFrom,
...
     {
-        pvd::PVField::shared_pointer raw(SerializationHelper::deserializeFull(payloadBuffer, transport.get()));
-        if(raw->getField()->getType()==pvd::structure) {
+        pvd::PVField::shared_pointer raw(
+                SerializationHelper::deserializeFull(payloadBuffer, transport.get()));
+        if (raw && raw->getField()->getType()==pvd::structure) {
             data = std::tr1::static_pointer_cast<pvd::PVStructure>(raw);
         } else {
-            // was originally possible, but never used
+            return; // ignore
```
anjohnson, line 327: Don't return here, passes a NULL down which is covered.

src/server/baseChannelRequester.cpp
```diff
@@ -66,7 +66,8 @@ int32 BaseChannelRequester::getPendingRequest()
 string BaseChannelRequester::getRequesterName()
 {
     std::stringstream name;
-    name << typeid(*_transport).name() << "/" << _ioid;
+    Transport& transport = *_transport;
+    name << typeid(transport).name() << "/" << _ioid;
```
anjohnson, line 70: Okay, but not hardening.

src/server/responseHandlers.cpp
```diff
@@ -1170,19 +1179,17 @@ void ServerChannelGetRequesterImpl::destroy()
...
-        if (_channelGet)
-        {
-            _channelGet->destroy();
-            _channelGet.reset();
-        }
+        channelGet.swap(_channelGet);
     }
+
+    if (channelGet)
+        channelGet->destroy();
```
anjohnson, line 1192: MAD checked and is good with this.

src/server/responseHandlers.cpp
```diff
@@ -1746,14 +1749,14 @@ void ServerChannelPutGetRequesterImpl::send(ByteBuffer* buffer, TransportSendCon
         else if ((QOS_GET_PUT & request) != 0)
         {
             ScopedLock lock(channelPutGet);
-            //Lock guard(_mutex);
+            Lock guard(_mutex);
             _pvPutBitSet->serialize(buffer, control);
             _pvPutStructure->serialize(buffer, control, _pvPutBitSet.get());
         }
```
anjohnson, line 1755: What data is being guarded here?

src/server/responseHandlers.cpp
```diff
@@ -2369,13 +2370,13 @@ void ServerChannelArrayRequesterImpl::send(ByteBuffer* buffer, TransportSendCont
     {
         if ((QOS_GET & request) != 0)
         {
-            //Lock guard(_mutex);
+            Lock guard(_mutex);
             ScopedLock lock(channelArray);
             _pvArray->serialize(buffer, control, 0, _pvArray->getLength());
```
anjohnson, line 2375: Locking order here is different than above QOS stuff, is this correct?

src/server/serverChannelImpl.cpp
```diff
@@ -103,8 +103,9 @@ void ServerChannel::printInfo() const
 
 void ServerChannel::printInfo(FILE *fd) const
 {
+    Channel& channel = *_channel;
     fprintf(fd,"CLASS        : %s\n", typeid(*this).name());
-    fprintf(fd,"CHANNEL      : %s\n", typeid(*_channel).name());
+    fprintf(fd,"CHANNEL      : %s\n", typeid(channel).name());
```
anjohnson, line 108: Ack, but not hardening.

src/remote/codec.cpp
```diff
@@ -242,6 +242,13 @@ void AbstractCodec::processReadNormal()  {
                         "not-a-first segmented message received in normal mode");
                 }
 
+                // reject negative payload size (sign-extends to a huge size_t)
+                if (_payloadSize < 0)
+                {
+                    invalidDataStreamHandler();
+                    throw invalid_data_stream_exception("negative payload size");
+                }
+
```
anjohnson, line 251: Remove this change, and the next one.

anjohnson (commented):
Core comments with detail fixes.

PR #227, opened 2026-07-28
Fixes for issues triggered by Windows CI

anjohnson:
Move `stopRequest()` calls in responseHandlers.cpp `RequesterImpl::send()` methods to before the payload gets serialized. See Matej's similar move and comments in the ChannelGet code when he was working on performance. The testChannelAccess test 148 fails with "other request pending" because of this ordering.

The factory stop method prevents crashes in the testCaProvider code during process shutdown.

### DIFF
--- a/src/ca/caProvider.cpp
+++ b/src/ca/caProvider.cpp
@@ -134,6 +134,12 @@ void CAClientFactory::start()
 
 void CAClientFactory::stop()
 {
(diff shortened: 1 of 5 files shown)


anjohnson:
Added another fix for requester callback diagnostics in testChannelAccess: Collect `cout` messages and output them using `testDiag()`.

